PR #6, opened 2019-10-31
Implements functioins from brqsort.c.

OmniBlade:


jeff-1amstudios:
this is great! where did the implementation come from? IDA / hand reversed / reimplemented from scratch?

OmniBlade:
IDA, no idea if it actually works, needs testing :D

jeff-1amstudios:
It would be nice to be able to unit test some of these funcs so we don’t get bogged down in random errors later on :)

IDA looks like it did a pretty good job though :)

There’s some plugin I saw where it will do a bulk conversion of all functions to C instead of one by one. How are you doing it at the moment?

OmniBlade:
It varies, if I can find a binary where it looks more or less spot on, then I'll rename the variables and call it done. If it still looks a mess I'll rewrite from the pseudo and drop down to asm if it looks like IDA missed something.

Yeah, we can start looking at unit testing. For simple stuff like this sort, we can probably just throw some data at it. For more complicated stuff we might want to hook against the original and run an original function side by side on the same data and compare the output. Did that with pathfinding in RA.

PR #7, opened 2019-10-31
Implements minimum changes to get BRender files to compile.

OmniBlade:
Not all of the changes will be strictly correct, but is the minimum I needed to do to get it to build the brender code on ubuntu 18.04.

### DIFF
--- a/src/BRSRC13/CORE/FW/diag.c
+++ b/src/BRSRC13/CORE/FW/diag.c
@@ -1,4 +1,6 @@
 #include "diag.h"
+#include "fwsetup.h"
+#include "printf.h"
 #include "stdlib.h"
 #include <stdarg.h>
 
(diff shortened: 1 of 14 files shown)


jeff-1amstudios:
this is a lot of work! :) I'm wary to merge it only because I'm still looking into a bug in the watcom debug dump tooling which could mean the headers need to be regenerated :/

OmniBlade:
It did look like some of the function signatures were incorrect, either variables had bled into the function parameters or random "unknown" fields had been added that didn't appear to be from the debug symbols that I could see. I've left the ones that successfully compiled even if I thought they were wrong, only changed blatantly incorrect ones.

jeff-1amstudios:
Why the changes to the `*cbfn` prototypes?  In the brender 1.2 sdk for example, it has 
```
typedef br_material * BR_CALLBACK br_material_find_cbfn(char *name);
```
which matches the prototypes in the generated headers

OmniBlade:
You are right, there was no need for me to make those changes to the typedefs, they are typedefing the function signature and declaring a pointer to it later which I must admit I'd never seen before and wasn't even sure it was valid C syntax. I've only ever seen typedefs used to create function pointer types.

OmniBlade:
You are right, there was no need for me to make those changes to the typedefs, they are typedefing the function signature and declaring a pointer to it later which I must admit I'd never seen before and wasn't even sure it was valid C syntax. I've only ever seen typedefs used to create function pointer types.

OmniBlade:
Okay, I've redone getting it to compile with as few changes as I could manage. Also fix a few issues with the previous code so it actually builds, whoops :)
Linking step fails due to duplicate symbols and one missing one which is in printf.c so I don't get why that isn't linking.

PR #10, opened 2019-11-16
Cleans up compile output, only half dozen warnings

jeff-1amstudios:
- Renames stdlib.h to avoid confusion and collision with stdlib
- Disables warnings for no return types and anon structs (for now)
- Fix linker issue with `BrSprintF` > `BrSprintf`


jeff-1amstudios:
Maybe renaming all the files in `CORE/STD` to be prefixed with `br` might be a good idea to keep in simple. Otherwise, need to figure out some compiler flags to make it work cleanly (doesn't seem to be happy in OSX at least with "stdlib.h" being included

jeff-1amstudios:
Hrm, maybe I wasn’t thinking. I guess some files just need <stdlib.h> and “stdlib.h”. I’ll fix up this PR assuming that works

jeff-1amstudios:
The reason I went down this path was because I was fixing “implicit definition of malloc” messages

PR #15, opened 2019-11-26
Implements BrFloat* functions.

OmniBlade:


test/Makefile
```diff
@@ -17,13 +17,14 @@ INC_DIRS := $(shell find $(SRC_DIR) -type d) $(BR_SRC_DIR) $(DR_SRC_DIR)
...
 CFLAGS ?= $(INC_FLAGS)
+LDFLAGS ?= -lm
 
 .PHONY: clean build run
 
 build: $(BUILD_DIR)/$(TARGET_EXEC)
 
 $(BUILD_DIR)/$(TARGET_EXEC): $(OBJS)
-	@$(CC) $(OBJS) -o $@ $(LDFLAGS)
+	@$(CC) $(OBJS) -o $@ $(LDFLAGS) -lm
```
jeff-1amstudios: Assume we don’t need this since it’s already in ldflags?

OmniBlade: Yeah, that was just there for testing, not sure how I let that slip by into the final commit.

PR #16, opened 2019-12-02
implements main method

jeff-1amstudios:


src/DETHRACE/watcom_functions.c
```diff
@@ -0,0 +1,28 @@
+#include "watcom_functions.h"
+
+#include <ctype.h>
+#include <libgen.h>
+#include <stdlib.h>
+#include <string.h>
+
+int stricmp(const char *a, const char *b) {
```
OmniBlade: stricmp is just another name for strcasecmp from posix <strings.h>. I would suggest adding a conditional define rather than rolling our own.

PR #24, opened 2020-05-16
Implements LoadModel

jeff-1amstudios:
- Implements BRender file + chunk loading code and the table used for decoding model files.
- Adds `NOT_IMPLEMENTED()` guards on all empty methods so when they get invoked unexpectedly we know about it immediately instead of getting random return values ;)
- Adds test for loading a simple model (https://github.com/jeff-1amstudios/dethrace/blob/bd98195d0b2cbbf5ea94781279f8b723f132ea3f/test/BRSRC13/test_v1dbfile.c)


src/BRSRC13/CORE/FW/datafile.c
```diff
@@ -87,11 +295,162 @@ br_uint_32 DfStructReadBinary(br_datafile* df, br_file_struct* str, void* base)
...
+            for (n = 0; n < 255; n++) {
+                c = BrFileGetChar(df->h);
+                if (!c || c == -1)
+                    break;
+                tmp_string[n] = c;
+            }
+            tmp_string[n] = 0;
+            tmp = BrResStrDup(df->res ? df->res : fw.res, tmp_string);
+            // TODO: how to make sure this works if sizeof(long) != sizeof(void*)
+            *(long*)mp = tmp;
```
OmniBlade, line 385: Probably best to use intptr_t from <stdint.h> which will be at least sizeof(void*) provided sufficient space is allocated in what mp points to?

src/BRSRC13/CORE/FW/datafile.c
```diff
@@ -440,18 +866,56 @@ int DfChunksInterpret(br_datafile* df, br_chunks_table* table) {
...
 
 // Offset: 13512
 // Size: 168
 int DfFileIdentify(br_uint_8* magics, br_size_t n_magics) {
-    static char text_magics[8];
-    static char binary_magics[8];
+    static char text_magics[8] = { '*', 'F', 'I', 'L', 'E', '_', 'I', 'N' };
+    static char binary_magics[8] = { '\0', '\0', '\0', '\x12', '\0', '\0', '\0', '\b' };
+
+    if (BrMemCmp(magics, text_magics, 8) == 0) {
```
OmniBlade, line 911: Use sizeof(magics) instead of hardcoded 8 for the comparisons?

PR #43, opened 2021-02-16
Provisional CMake based build system.

### LINKED ISSUE #5: Suggestion: Use a meta build system.
A meta build system will allow a more diverse range of platforms to build the project, the main one being MSVC which doesn't support make files. My preference would be for cmake as it has wide support.

OmniBlade:
A CMake based build system makes it a bit easier to develop on other platforms, specifically windows and msvc, but could be used to more easily target other platforms as well.

Fixes #5


OmniBlade:
Looks like to support modernish OpenGL on msvc a loader library will be needed, something like GLEW or glad? Thoughts?

src/DETHRACE/common/graphics.c
```diff
@@ -1030,8 +1030,8 @@ void DRPixelmapRectangleOnscreenCopy(br_pixelmap* pDest, br_int_16 pDest_x, br_i
 
     source_row_wrap = pSource->row_bytes - pWidth;
     dest_row_wrap = pDest->row_bytes - pWidth;
-    dest_ptr = pDest->pixels + (pDest->row_bytes * pDest_y + pDest_x);
-    source_ptr = pSource->pixels + (pSource->row_bytes * pSource_y + pSource_x);
+    dest_ptr = (char*)pDest->pixels + (pDest->row_bytes * pDest_y + pDest_x);
```
jeff-1amstudios, line 1033: most of these are `tU8*` instead of char, which will avoid `-Wpointer-sign` warnings

jeff-1amstudios:
This is great, thanks for getting this off the ground!

Where does `miniposix` come from? I don't see any obvious attribution?

I tried the build of this branch, I get linker errors
```
[100%] Linking C executable dethrace
Undefined symbols for architecture x86_64:
  "_BrFontFixed3x5", referenced from:
      _InitHeadups in displays.c.o
  "_BrFontProp7x9", referenced from:
      _InitHeadups in displays.c.o
  "__BrDefaultDiagHandler", referenced from:
      _BrFwBegin in libbrender.a(fwsetup.c.o)
  "__BrDefaultModel", referenced from:
      _BrV1dbBegin in libbrender.a(dbsetup.c.o)
```

OmniBlade:
I cobbled miniposix together myself, a good chunk of it is based on https://github.com/win32ports, while the fnmatch is ported from freebsd and the getopt implementation is a public domain implementation, though I don't recall where I picked that up from specifically. The copyright information provided with the originals is preserved in the source files.

Not sure where the linking issue is coming from, I tested building in MSVC 2019 and on the WSL Ubuntu 18.04 distro and it built and linked successfully on both platforms via the cmake generated build files.

src/DETHRACE/common/graphics.c
OmniBlade: Thanks, I'll address that.

OmniBlade: Casts should match the pointers they are getting assigned to now.

jeff-1amstudios:
Ok thanks, I’ll have more of a look into the linker issue

jeff-1amstudios:
confirmed - works fine on Linux. The linker error is only present on OSX. I'll test the simplest cmake example (https://cmake.org/examples/) and add some global vars to the example library to try and repro

OmniBlade:
I've added workflows to test the build on the 3 main platforms and see the same linking issues for macOS. There don't seem to be any compiler warnings about those symbols specifically when compiling those sources. You do have the `-fno-pie` linker flag specified in your make files which I don't think I reproduced in the CMake file, but other than that I'm not sure what would be different between the two build systems.

OmniBlade:
All CI's build now and the linux and macOS ones run the unit tests. Some tests fail on windows due to differences in the path to the temp folder, so I don't run them there yet.

jeff-1amstudios (approved):

PR #65, opened 2021-12-29
Fix MSVC tests + add mingw support

madebr:
This PR fixes various problems with buiding dethrace with mingw and MSVC.
It also enables running the Windows tests on github actions.

I've documented the reasoning behind each commit in the commit messages.

- This is the result of github actions: https://github.com/madebr/dethrace/actions/runs/1633319741
- I've sent the cglm patch to upstream: https://github.com/recp/cglm/pull/234


test/BRSRC13/test_resource.c
```diff
@@ -62,17 +62,21 @@ void test_resource_BrResFree_Child() {
     header = (resource_header*)(((char*)child) - sizeof(resource_header));
     TEST_ASSERT_EQUAL_INT(0xDEADBEEF, header->magic_num);
     BrResFree(r);
+
+#ifndef _DEBUG
```
dethrace-labs, line 66: why do we need to add this guard?

test/DETHRACE/test_controls.c
```diff
@@ -24,7 +24,7 @@ void test_controls_CheckKevKeys() {
         gKeys_pressed = 0;
         result = KevKeyService();
     }
-    sleep(2);
+    sleep_s(2);
```
dethrace-labs, line 27: what is `sleep_s`?

dethrace-labs: nevermind, I see

test/BRSRC13/test_resource.c
madebr: MSVC in debug mode overwrites all free'd memory with some mask to detect use-after-free.
`BrResFree(r)` will also free all its children, so we cannot access the `header` pointer anymore.

dethrace-labs (approved):

test/BRSRC13/test_resource.c
dethrace-labs: ah, got it

dethrace-labs:
thanks for this! I didn't know about `UnitySetTestFile` - I've been ignoring the reported file name in test errors because it was always reporting `main.c` :)

PR #66, opened 2022-01-03
Implement all brender matrix functions

madebr:
This pr adds implementations for all matrix related functions.
It also tests each function at least once.

fyi, the rolling ball algorithm is described in this paper:
http://citeseerx.ist.psu.edu/viewdoc/summary;jsessionid=FC5BB3EADECE14753EDF5C15AC6A1346?doi=10.1.1.69.2209
(equation 3 with nz = 0)


dethrace-labs (approved):

madebr:
@dethrace-labs 
I think you need to enable actions for me first before merging.

PR #95, opened 2022-02-27
Implements depth effects

dethrace-labs:
- Implements Fog and Darkness depth effects via shadetables
- Discards transparent pixels to fix rendering black pixels instead of transparent pixels
Screen Shot 2022-03-01 at 2 17 40 PM


src/DETHRACE/common/graphics.c
```diff
@@ -1188,10 +1188,11 @@ int ConditionallyFillWithSky(br_pixelmap* pPixelmap) {
     int bgnd_col;
     LOG_TRACE("(%p)", pPixelmap);
 
-    if (gProgram_state.current_depth_effect.sky_texture
-        && (!gLast_camera_special_volume || gLast_camera_special_volume->sky_col < 0)) {
-        return 0;
-    }
+    // TODO: Uncomment when ExternalSky is fully implemented
```
madebr, line 1191: Also add a `WARN_ONCE`. That way it remains visible.

dethrace-labs: good call, done

PR #96, opened 2022-02-27
Truckload of updates

madebr:
Highlights:
- blip in the map is working:
Screenshot from 2022-02-27 22-43-12
- save + load screens + logic "working".
I don't guarantee any compatibility with the original carmageddon savegames.
So create savegames before using it...
Screenshot from 2022-02-28 00-32-28
- APO keys working (and not working in single player campaign):
Screenshot from 2022-02-28 00-37-57
- Various implementations

```
Overall progress:
55% of all functions implemented. (1900 completed / 3471 total)
```


dethrace-labs:
While this is awesome work, as usual(!), this kind of PR is a little difficult for me to handle. 

PRs that implement a feature or behavior end to end are easier to 
a) test before merging
b) because we can test them, that ensures all the code going in to `main` is working correctly as best as we can tell.

This PR has some things that should be easy to merge:
- Load / Save full end to end
- Map rendering (enough to know that what you've done so far is working)

Things that I don't mind as extras in the PR:
- Extra BRender functions that might/might not be required specifically for dethrace
- Tidying (replacing 0's with `NULL`, adding `STUB` calls where appropriate, removing incorrectly inlined function calls etc etc

Things that make it hard to merge:
- Functions implemented which either aren't called yet, or are part of a feature that is not completed end-to-end, so I have no way to sanity check except by looking at each function myself:
- Examples: `StripBlendedFaces`, the crush code, the sparks code, the shrapnel code, flame code, smoke code, etc.
- `MakeCarIt` is a good example of what I'm trying to say - that code is not going to be used until we have network play. When we start on network play, that would be a good time to implement it, because it will get used and tested. Right now it has an implementation which we don't know is right or not. As we add functionality, I want to be able to trust the code that is already there. It makes debugging so much nicer when we don't hit bugs in a function that was implemented earlier and was never validated.

I'm torn, because I see all your good work, but I want to get into the habit of merging full features without so much random extras

src/DETHRACE/common/brucetrk.c
```diff
@@ -103,8 +153,42 @@ intptr_t FindNonCarsCB(br_actor* pActor, tTrack_spec* pTrack_spec) {
...
     LOG_TRACE("(%p, %p)", pActor, pTrack_spec);
 
-    STUB();
-    return 0;
+    if (pActor->identifier != NULL && pActor->identifier[0] == '&' && pActor->identifier[1] >= '0' && pActor->identifier[1] <= '9') {
+        i = (pActor->identifier[4] - '0') * 1000 + (pActor->identifier[5] - '0') * 100 + (pActor->identifier[6] - '0') * 10 + (pActor->identifier[7] - '0');
+        if (i < 0 || pTrack_spec->ampersand_digits <= i) {
+            return 1;
+        }
+        r1 = BR_SQR3(pActor->t.t.mat.m[0][0], pActor->t.t.mat.m[0][1], pActor->t.t.mat.m[0][2]);
```
dethrace-labs, line 161: I normally use `BrVector3LengthSquared` here - probably `BR_SQR3` could be removed from the public brender include

madebr: I used this one because I found it weird to cast a row of a matrix to a vector, but it makes sense to do it.
So I'll do that from now on.

PR #97, opened 2022-03-03
Don't grab the mouse when debugging

madebr:
This pr adds 2 main features to make our life easier:
- don't grab the mouse when a debugger is attached.
This means that, when debugging in an IDE, the mouse won't be hidden.
Right now, the mouse is hidden when some breakpoint is hit which is annoying.
- do `abort()` instead of `exit(1)` on a panic when a debugger is attached,
such that you are thrown right to the correct spot when e.g. an unimplemented function is encountered.

I copied the code to detect a debugger on Macos from this apple faq page: https://developer.apple.com/library/archive/qa/qa1361/_index.html
There is an important note about the code not working on a release build.
Since I don't own an apple (and you do), can you take a look at this and fix it?
You probably might have to `#ifdef _DEBUG` this function and return `0` when building a release binary.

This is another way to detect a debugger on Macos
https://zgcoder.net/ramblings/osx-debugger-detection/#mach-exception-handlers

Since gdb is not really used by apply, perhaps this 2nd method should be used instead.

### DIFF
--- a/src/BRSRC13/CORE/V1DB/v1dbfile.c
+++ b/src/BRSRC13/CORE/V1DB/v1dbfile.c
@@ -969,7 +969,7 @@ int FopRead_TRANSFORM(br_datafile* df, br_uint_32 id, br_uint_32 length, br_uint
         }
     }
     if (t == 7) {
-        LOG_PANIC("transform type not found!")
+        LOG_PANIC("transform type not found!");
     }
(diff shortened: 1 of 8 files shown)


dethrace-labs:
Seems to work well enough for me - I can build in release mode and it happily runs with no debugger attached. I don't really use a debugger, so I'd be happy enough if the osx one is stubbed with `return 0;`, but happy to merge as-is for now

dethrace-labs (approved):

PR #103, opened 2022-03-11
Start checking checkpoints + laps

madebr:
- Count checkpoints + checkpoints
- Fix flashing checkpoints in map mode
- Free taken pixelmap when leaving the change car dialog

Screenshot from 2022-03-11 15-34-43

(I wasn't able to get a screenshot of the race completed fancy headup)
Screenshot from 2022-03-11 15-38-50
Screenshot from 2022-03-11 15-38-55

### DIFF
--- a/src/DETHRACE/common/structur.c
+++ b/src/DETHRACE/common/structur.c
@@ -5,6 +5,7 @@
 #include "cutscene.h"
 #include "displays.h"
 #include "drmem.h"
+#include "finteray.h"
 #include "flicplay.h"
 #include "globvars.h"
 #include "globvrkm.h"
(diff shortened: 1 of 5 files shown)


madebr:
The flashing overlay of the 2nd checkpoint is offset with 1 pixel.
Is this a bug or also present with the original version?

Screenshot from 2022-03-11 17-19-05

RebelSloth:
The 2nd checkpoint glitch is present in the original DOS version (captured with DOSBox using the original CDs)
carma

src/DETHRACE/common/structur.c
```diff
@@ -128,14 +129,46 @@ void RaceCompleted(tRace_over_reason pReason) {
...
+        return;
+    }
+    gLast_checkpoint_time = GetTotalTime();
+    if (gCheckpoint < gCheckpoint_count) {
+        gCheckpoint++;
+    } else {
+        gCheckpoint = 1;
+        gLap++;
+        if (gLap == gTotal_laps) {
+            PratcamEvent(33);  // FIXME: or PratcamEventNow
```
jeff-1amstudios, line 156: What does the comment mean about PratcamEventNow?

madebr: I'm not sure this function is PratcamEvent or PratcamEventNow.
It's probably PratcamEvent, but hence the comment

dethrace-labs: I agree, its `PratcamEvent`. The DOS version is the most closely related to the symbol dump for ordering of functions and globals, confirmed with that

dethrace-labs (approved):

PR #106, opened 2022-03-15
Visual Studio Fixes

madebr:
- Fixes runtime errors when using debug runtime (`-MDd`/`-MTd`)
- Add 2 cmake options (`DETHRACE_IDE_ROOT_DIR` and  `DETHRACE_IDE_ARGUMENTS`) to enable debugging within MSVC.
- Fix the race timer counting down too fast when not frame limiting: the game would floor the frame period to 10ms.

Other fixes:
- fix bug in `ClearWobbles` + typo in `LoadGeneralParameters`
- use floats instead of doubles in varous places
- compare against `NULL`s
- stub `ChangeAmbientPratcamNow` so the game doesn't abort anymore when finishing/timing out.

### DIFF
--- a/src/DETHRACE/common/mainloop.c
+++ b/src/DETHRACE/common/mainloop.c
@@ -389,7 +389,7 @@ void UpdateFramePeriod(tU32* pCamera_period) {
         }
     } else {
         // The following makes the timer go too fast when the real frame rate is high (=low frame period)
-#ifndef BRENDER_FIX_BUGS
+#ifndef DETHRACE_FIX_BUGS
         gFrame_period = 10;
 #endif
         gLast_tick_count = new_tick_count;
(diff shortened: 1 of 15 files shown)


src/DETHRACE/common/mainloop.c
```diff
@@ -388,7 +388,10 @@ void UpdateFramePeriod(tU32* pCamera_period) {
             gLast_tick_count = new_tick_count;
         }
     } else {
+        // The following makes the timer go too fast when the real frame rate is high (=low frame period)
+#ifndef BRENDER_FIX_BUGS
```
dethrace-labs, line 392: `DETHRACE_FIX_BUGS`?

dethrace-labs: the same behavior is also applied to the camera on line 399 `*pCamera_period = 10;` - should that be removed also?

madebr: Perhaps, I think the camera_period is only used by the action replay.
So that's something we need to remind when we start touching that.

dethrace-labs (approved):

PR #109, opened 2022-03-21
Start To Mouse

### LINKED ISSUE #75: Crash when selecting "View Racers" in the "Start Race" menu
Managed to run into following crash when in the "Start Race" menu selected "View Info" -> "View Racers" button

Stack trace:
0: MemReadU32 at dethrace/src/DETHRACE/common/loading.c:220
1: StartFlic at dethrace/src/DETHRACE/common/flicplay.c:689
2: ChangePanelFlic at dethrace/src/DETHRACE/common/flicplay.c:1730
3: SetOpponentFlic at dethrace/src/DETHRACE/common/racestrt.c:386
4: DrawSceneyMappyInfoVieweyThing at dethrace/src/DETHRACE/common/racestrt.c:397 
5: StartRaceGoAhead at dethrace/src/DETHRACE/common/racestrt.c:447
6: DoInterfaceScreen at dethrace/src/DETHRACE/common/intrface.c:413
7: DoSelectRace at dethrace/src/DETHRACE/common/racestrt.c:804
8: DoGame at dethrace/src/DETHRACE/common/structur.c:319
9: DoProgram at dethrace/src/DETHRACE/common/structur.c:506
10: GameMain at dethrace/src/DETHRACE/common/main.c:96
11: original_main at dethrace/src/DETHRACE/pc-dos/dossys.c:690
12: main at dethrace/src/DETHRACE/main.c:30

Happens with current master ba4b7630fbb

madebr:
This pr:
- adds mouse support
- draws the cursor
- draws giblets (=blood drops)

Screenshot from 2022-03-22 00-35-43

Fixes #75

### DIFF
--- a/src/DETHRACE/pc-dos/dossys.c
+++ b/src/DETHRACE/pc-dos/dossys.c
@@ -532,13 +532,12 @@ void PDMouseButtons(int* pButton_1, int* pButton_2) {
     br_int_32 mouse_y;
     LOG_TRACE("(%p, %p)", pButton_1, pButton_2);
 
-    STUB_ONCE();
-    *pButton_1 = 0;
-    *pButton_2 = 0;
+    Harness_Hook_GetMouseButtons(pButton_1, pButton_2);
 }
 
 // IDA: void __usercall PDGetMousePosition(int *pX_coord@<EAX>, int *pY_coord@<EDX>)
 void PDGetMousePosition(int* pX_coord, int* pY_coord) {
+#if 0
     br_uint_32 mouse_buttons;
     br_int_32 mouse_x2;
     br_int_32 mouse_y2;
@@ -546,8 +545,15 @@ void PDGetMousePosition(int* pX_coord, int* pY_coord) {
     int delta_y;
     static br_int_32 mouse_x;
     static br_int_32 mouse_y;
+#endif
+    int w;  // Added by DethRace
+    int h;  // Added by DethRace
     LOG_TRACE("(%p, %p)", pX_coord, pY_coord);
-    NOT_IMPLEMENTED();
+
+    Harness_Hook_GetWindowSize(&w, &h);
+    Harness_Hook_GetMousePosition(pX_coord, pY_coord);
+    *pX_coord = *pX_coord * gGraf_data[gGraf_data_index].width / w;
+    *pY_coord = *pY_coord * gGraf_data[gGraf_data_index].height / h;
 }
 
 // IDA: int __cdecl PDGetTotalTime()
--- a/src/DETHRACE/pc-dos/dossys.c
+++ b/src/DETHRACE/pc-dos/dossys.c
@@ -550,10 +550,7 @@ void PDGetMousePosition(int* pX_coord, int* pY_coord) {
     int h;  // Added by DethRace
     LOG_TRACE("(%p, %p)", pX_coord, pY_coord);
 
-    Harness_Hook_GetWindowSize(&w, &h);
     Harness_Hook_GetMousePosition(pX_coord, pY_coord);
-    *pX_coord = *pX_coord * gGraf_data[gGraf_data_index].width / w;
-    *pY_coord = *pY_coord * gGraf_data[gGraf_data_index].height / h;
 }
 
 // IDA: int __cdecl PDGetTotalTime()
--- a/src/DETHRACE/pc-dos/dossys.c
+++ b/src/DETHRACE/pc-dos/dossys.c
@@ -537,7 +537,6 @@ void PDMouseButtons(int* pButton_1, int* pButton_2) {
 
 // IDA: void __usercall PDGetMousePosition(int *pX_coord@<EAX>, int *pY_coord@<EDX>)
 void PDGetMousePosition(int* pX_coord, int* pY_coord) {
-#if 0
     br_uint_32 mouse_buttons;
     br_int_32 mouse_x2;
     br_int_32 mouse_y2;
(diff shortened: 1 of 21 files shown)


src/harness/os/linux.c
```diff
@@ -285,3 +284,28 @@ void OS_InstallSignalHandler(char* program_name) {
...
+    if (f != NULL) {
+        return f;
+    }
+    char buffer[512];
+    char buffer2[512];
+    strcpy(buffer, pathname);
+    strcpy(buffer2, pathname);
+    char* pDirName = dirname(buffer);
+    char* pBaseName = basename(buffer2);
+    DIR* pDir = opendir(pDirName);
```
dethrace-labs, line 299: if we re-write this using `OS_FindFirstFile` and `OS_FindNextFile`, is there any reason we couldn't move this functionality into `Harness_Hook_fopen`? If thats possible, that would be nice, since that would be the single place for getting a `FILE *` handle. Loading a file from an archive or straight from an http request etc belong in a non-OS specific place.

for example:
```
FILE* f = fopen(pathname, mode);
if (f != NULL) {
    return f;
}
...
char* file_found = OS_GetFirstFileInDirectory(pathname);
    while (file_found != NULL) {
        if (strcasecmp(pBaseName, file_found) == 0) {
            strcat(pDirName, "/");
            strcat(pDirName, file_found);
            return fopen(pDirName, mode);
       }
       file_found = OS_GetNextFileInDirectory(pathname);
    }
```

dethrace-labs: Also, MacOS _can_ run with a case sensitive filesystem (https://coderwall.com/p/mgi8ja/case-sensitive-git-in-mac-os-x-like-a-pro), so the same code that currently exists here in `linux.c` would otherwise need to duplicated into `macos.c`. Maybe irrelevant but Haiku OS is also case sensitive if we ever want to support it

madebr: We can also move this case insensitive search to a separate file + header such that different operating systems can make use of it.

src/harness/harness.c
```diff
@@ -264,13 +268,34 @@ void Harness_Hook_BrBufferUpdate(br_pixelmap* pm, br_token use, br_uint_16 flags
...
 
 void Harness_Hook_S3StopAllOutletSounds() {
 }
 
-void Harness_Hook_FlushRenderer() {
-    renderer->FlushBuffers(last_colour_buffer, last_depth_buffer);
+// Various hooks
+void Harness_Hook_Delay_ms(int ms) {
+    Time_Delay_ms(ms);
+}
```
madebr, line 296: I noticed there is also `void OS_Sleep(int delay_ms)`.

src/harness/harness.c
```diff
@@ -264,13 +268,34 @@ void Harness_Hook_BrBufferUpdate(br_pixelmap* pm, br_token use, br_uint_16 flags
     }
 }
 
+// Input hooks
+void Harness_Hook_GetMousePosition(int* pX, int* pY) {
```
dethrace-labs, line 272: What if we made this `Harness_Hook_GetRelativeMousePosition`, and has this function to the translation to window co-ordinates? Then we could avoid exposing also `Harness_Hook_GetWindowSize`?

src/harness/harness.c
dethrace-labs: yes, lets use `OS_Sleep` instead. Also, I don't think we need to have a harness_hook to front an `OS_*` call - we can call `Harness_*` or `OS_*` functions direct from OG code

src/DETHRACE/pc-dos/dossys.c
```diff
@@ -532,22 +532,28 @@ void PDMouseButtons(int* pButton_1, int* pButton_2) {
...
     static br_int_32 mouse_y;
+#endif
+    int w;  // Added by DethRace
+    int h;  // Added by DethRace
     LOG_TRACE("(%p, %p)", pX_coord, pY_coord);
-    NOT_IMPLEMENTED();
+
+    Harness_Hook_GetWindowSize(&w, &h);
+    Harness_Hook_GetMousePosition(pX_coord, pY_coord);
+    *pX_coord = *pX_coord * gGraf_data[gGraf_data_index].width / w;
```
dethrace-labs, line 555: what is the OG code expecting the value of `pX_coord` to be? `SDL_GetMouseState` already gets the position relative to the window., and the OG code looks like that is what the OG `PDGetMousePosition` is also doing. Why do we need this calculation?

madebr: These calls are not needed anymore. The scaling factor is stored within sdl_gl.c.

src/harness/io_platforms/io_platform.h
```diff
@@ -6,7 +6,10 @@
 tRenderer* Window_Create(char* title, int width, int height, int pRender_width, int pRender_height);
 void Window_PollEvents(void);
 void Window_Swap(int delay_ms_after_swap);
+void Window_GetSize(int* pW, int* pH);
```
dethrace-labs, line 9: should we remove this now?

src/DETHRACE/pc-dos/dossys.c
```diff
@@ -532,22 +532,25 @@ void PDMouseButtons(int* pButton_1, int* pButton_2) {
...
 void PDGetMousePosition(int* pX_coord, int* pY_coord) {
+#if 0
     br_uint_32 mouse_buttons;
     br_int_32 mouse_x2;
     br_int_32 mouse_y2;
     int delta_x;
     int delta_y;
     static br_int_32 mouse_x;
     static br_int_32 mouse_y;
+#endif
```
dethrace-labs, line 548: we can remove this now? 👍

madebr: Yup can be removed.

src/harness/io_platforms/io_platform.h
madebr: Yes, the function is missing.

dethrace-labs (approved):

PR #118, opened 2022-04-06
Add pedestrians + power-ups

madebr:
This PR:
- Adds pedestrians + power-ups
- Implements some power-ups, such as the instant handbrake
  (But it behaves strangely when the car is in the air: the car slowly moves back to earth)
- Implements the `FlyCar` cheat (cheat code is `molemode`). The car control remains awkward, imho.
- Add the pedestrian edit functions (`ibetyoucantprintcunt` -> `F4` until PEDESTRIANS)
- Adds the damage edit functions (`ibetyoucantprintcunt` -> `F4` until DAMAGE appears -> smash <kbd>0</kbd> to <kbd>6</kbd> or <kbd>SHIFT</kbd>+<kbd>0</kbd>-<kbd>SHIFT</kbd>+<kbd>6</kbd>.
- Modify various functions to use floats and the brender macro's. There should be no functional change  (I changed something in `CollideCamera2` though, which I think was misplaced. Please verify this.)
- Fix an undefined behavior warning in flicplay: it tried to read/write a u16 from/to an unaligned memory address. I fixed this by using an intermediate helper function that does a `memcpy`.
- Fix the pedestrian/powerup shown on maps powerup (use integers instead of floats, when calculating the pixel locations)
- Implement some functions to allow changing the graphics quality (=see options -> graphics panel)
- Modify the progress script to give more attention to the number of missing functions, instead of the number of functions already implemented (the previous sort order is still available through `--sort percent`).

The state after this PR is as follows:
```
name                                          percent completed / total missing
=======================================
src/DETHRACE/common/world.c                       40%       100 /   252   12.49
src/DETHRACE/common/network.c                     13%        13 /   103    7.40
src/DETHRACE/common/opponent.c                    35%        48 /   137    7.31
src/DETHRACE/common/piping.c                      40%        49 /   123    6.08
src/DETHRACE/common/car.c                         58%        94 /   161    5.51
src/DETHRACE/common/newgame.c                     19%        14 /    73    4.85
src/DETHRACE/pc-dos/dosnet.c                       2%         1 /    49    3.94
src/DETHRACE/pc-dos/dossys.c                      38%        29 /    77    3.94
src/DETHRACE/common/netgame.c                     17%         9 /    52    3.53
src/DETHRACE/common/controls.c                    69%        84 /   121    3.04

...
name                                          percent completed / total missing
=======================================
total                                            65%      2263 / 3480
```

The top files are the biggest contributors of the missing functionality.


src/DETHRACE/common/car.c
```diff
@@ -5054,65 +4947,36 @@ int CollideCamera2(br_vector3* car_pos, br_vector3* cam_pos, br_vector3* old_cam
...
+                        BrVector3Scale(&b, &face_list[0].normal, sc);
+                        BrVector3Sub(&face_list[1].normal, &face_list[1].normal, &b);
+                        BrVector3Scale(&tv2, &face_list[1].normal, hither - ts2);
+                        BrVector3Accumulate(cam_pos, &tv2);
                     }
                 }
             }
         }
-        i += k;
+        i += k + 1;
```
dethrace-labs, line 4979: where did this `+1` come from? I don't see that in mine?

dethrace-labs: which is the part of this function that you thought might have been misplaced?

dethrace-labs:
Its very cool! I do have some comments after playing this branch and comparing it to the OG - 

 - the peds running / hands waving animation is much faster than the original - the sprites look flickery as they run. It gets better with a `--fps=20` limiter, so there might be some timing fixes we need to do to base incrementing the animation frame based on elapsed time, rather than just whenever the screen is rendered (See what I did with the animated flames - I think it might be the same problem).
 - the death animation, on the other hand, seems too _slow_ - definitely noticeable compared to OG

src/DETHRACE/common/car.c
```diff
@@ -5035,17 +4932,13 @@ int CollideCamera2(br_vector3* car_pos, br_vector3* cam_pos, br_vector3* old_cam
                         tv.v[2] = tv2.v[2] + tv.v[2];
                     }
                 }
+                BrVector3Add(cam_pos, car_pos, &tv);
             }
-            cam_pos->v[0] = car_pos->v[0] + tv.v[0];
-            cam_pos->v[1] = car_pos->v[1] + tv.v[1];
-            cam_pos->v[2] = car_pos->v[2] + tv.v[2];
```
madebr, line 5041: @dethrace-labs 
This little snippet.

src/DETHRACE/common/car.c
madebr: The original code was correct.
The `+1` comes from the `for` post-increment.
I was looking at a `while` implementation of the loop which combined the `+k` and `+1`.

src/DETHRACE/common/car.c
dethrace-labs: thanks for pointing it out. I agree, you have it it in the right place 👍

src/DETHRACE/common/piping.c
```diff
@@ -75,7 +75,8 @@ int PipeSearchForwards() {
 int IsActionReplayAvailable() {
     LOG_TRACE("()");
 
-    return gPipe_buffer_start != NULL;
+    STUB_ONCE();
```
dethrace-labs, line 78: did you mean to re-stub this?

src/DETHRACE/common/powerup.c
```diff
@@ -1211,3 +1304,28 @@ void LoseAllLocalPowerups(tCar_spec* pCar) {
         }
     }
 }
+
+// Added by dethrace
+void GetPowerupMessage(int pN, char* pMessage) {
```
dethrace-labs, line 1309: why did we add this function?

src/DETHRACE/common/raycast.c
```diff
@@ -253,9 +243,9 @@ int DRScenePick2D(br_actor* world, br_actor* camera, dr_pick2d_cbfn* callback, v
     DRActorToRoot(camera, world, &camera_tfm);
     BrMatrix34Inverse(&gPick_model_to_view__raycast, &camera_tfm);
     scale = cos(BrAngleToRadian(camera_data->field_of_view / 2));
-    scale = scale / sin(scale);
+    scale = scale / sin(BrAngleToRadian(camera_data->field_of_view / 2));
```
dethrace-labs, line 246: this change doesn't look correct based on my looking of the OG game?

dethrace-labs: hmm, nevermind, looking at the assembly you are right, my tooling shows the wrong thing!

src/harness/renderers/gl/shaders.c
```diff
@@ -50,6 +50,7 @@ const char* fs_3d = "#version 330 core\n"
                     "\n"
                     "out uint palette_index;\n"
                     "\n"
+                    "uniform vec3 pixels_transform[2];\n"
```
dethrace-labs, line 53: why not store this as a `mat3` here instead of 2 `vec3`s?

src/harness/renderers/gl/stored_context.h
```diff
@@ -18,6 +18,7 @@ typedef struct tStored_material {
     br_uint_32 flags;
     br_pixelmap* shade_table;
     br_pixelmap* pixelmap;
+    br_vector3 map_transform[2];
```
dethrace-labs, line 21: why not a `br_matrix23` ?

src/DETHRACE/common/car.c
dethrace-labs: Agree, you have this in the right place.

src/harness/renderers/gl/shaders.c
madebr: A texture transformation is stored as a br_matrix23 (=6item), so I thought to be smart and avoid sending the extra (0,0,1) vector to the GPU.

src/harness/renderers/gl/stored_context.h
madebr: Indeed, it can be a `br_matrix23` here

src/DETHRACE/common/piping.c
madebr: No, this is an accidental change.
Thanks for pointing it out.

src/harness/renderers/gl/shaders.c
dethrace-labs: true, it might be a little less data , but then on the other hand you then have to construct a new `mat3` for every texel that the shader processes ;)

madebr: In the shader I initialize the `mat3` with zeros, but my thought process was that the optimizer will be able to optimize out 0 multiplications.
On the other hand, sending these 8 bytes will not be a performance killer.

dethrace-labs:
In this branch, theres a weird shaking in the camera when the car is close to a wall and you pan the camera around so that it has to avoid colliding with the wall, its not present in `main`. I'll try and track it down. The ped behavior though now looks good to me - maybe one of the subsequent changes fixed something, or maybe I just thought I saw some unexpected behavior last time!

madebr:
Thanks for taking care of this one!

PR #126, opened 2022-05-01
Flush buffers in EndScene

### LINKED ISSUE #125: Incorrect HUD rendering in races without depth effects
In races like "Fridge Racer", the race HUD can be obstructed by the 3d scene.
Screen Shot 2022-05-02 at 8 44 14 AM

dethrace-labs:
- Moves `FlushBuffers` to `EndScene` instead of `BrPixelmapDoubleBuffer`. In races without depth effects, flushing the 3d scene was happening _after_ 2d rendering, causing the 3d scene to appear in front of the HUD
- Fixes #125 

Screen Shot 2022-05-02 at 8 47 42 AM


madebr:
Small patch, big difference!
Thanks!

PR #130, opened 2022-05-30
Audio support

dethrace-labs:
- Adds audio support, courtesy of [miniaudio](https://miniaud.io/)
- Fixes smack videos not clearing screen between frames
- Fixes pedestrians running towards the car instead of away


mackron:
FYI, I went ahead and gave this a quick try. Just a few things I found:
- In sound.c, the `UsePathFileToDetermineIfFullInstallation()` function uses fixed sized arrays of capacity 80 which is too small. My application path is this, and it just breaks with a stack corruption error: "C:\Program Files (x86)\GOG Galaxy\Games\Carmageddon Max Pack\CARSPLAT/DATA"
- Same as above, but in `S3LoadSample()` in s3audio.c.
- In audio.c and s3audio.c, the `fopen(pThe_path, "r");` calls need to be changed to `fopen(pThe_path, "rb");` (note the change from "r" to "rb").

After I fixed the above issues it seems to work quite nicely on Windows for me. Good work!

PR #135, opened 2022-08-23
Big endian fixes

BSzili:
This could be useful for porting Dethrace to XB360, PS3, Wii consoles, or perhaps more obscure platforms.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -17,6 +17,15 @@ option(BUILD_TESTS "Build unit tests." OFF)
 
 find_package(SDL2 REQUIRED)
 
+include(TestBigEndian)
+test_big_endian(IS_BIGENDIAN)
+
+if (IS_BIGENDIAN)
+    add_compile_definitions(BR_ENDIAN_BIG=1)
+else()
+    add_compile_definitions(BR_ENDIAN_LITTLE=1)
+endif()
+
 add_subdirectory(lib/libsmacker)
 add_subdirectory(lib/glad)
 add_subdirectory(lib/miniaudio)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,16 +15,10 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR})
 
 option(BUILD_TESTS "Build unit tests." OFF)
 
-find_package(SDL2 REQUIRED)
-
 include(TestBigEndian)
 test_big_endian(IS_BIGENDIAN)
 
-if (IS_BIGENDIAN)
-    add_compile_definitions(BR_ENDIAN_BIG=1)
-else()
-    add_compile_definitions(BR_ENDIAN_LITTLE=1)
-endif()
+find_package(SDL2 REQUIRED)
 
 add_subdirectory(lib/libsmacker)
 add_subdirectory(lib/glad)
--- a/src/BRSRC13/CMakeLists.txt
+++ b/src/BRSRC13/CMakeLists.txt
@@ -37,6 +37,12 @@ if(BRENDER_FIX_BUGS)
     target_compile_definitions(brender PRIVATE BRENDER_FIX_BUGS)
 endif()
 
+if(IS_BIGENDIAN)
+    target_compile_definitions(brender PRIVATE BR_ENDIAN_BIG=1)
+else()
+    target_compile_definitions(brender PRIVATE BR_ENDIAN_LITTLE=1)
+endif()
+
 target_sources(brender PRIVATE
     include/brender/brender.h
     include/brender/br_types.h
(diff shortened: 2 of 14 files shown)


classilla:
Yes, *please* don't just `defined(__ppc__)` -- there are little-endian PowerPC systems too. Raptor POWER9s can swing both ways (my Talos II is little-endian).

BSzili:
It looks like TestBigEndian is deprecated, can CMAKE_C_BYTE_ORDER be used here? The save game code should be easy to fix, I'll make SWAP32_BE a no-op if BR_ENDIAN_BIG is 1.

madebr:
That's fine to me.
Supporting older CMake versions is not a priority here.

@dethrace-labs. WDYT? CMake 3.20 is widely available by now.

Don't forget to [bump the the minimum required version](https://github.com/dethrace-labs/dethrace/blob/05f16d3066ce8eda39727cf10de0b06688d2f061/CMakeLists.txt#L1).

BSzili:
Just to clarify I'm not very familiar with CMake, so I don't insist on using the newer method. I'll look into whichever way you say would suit the project better.

dethrace-labs:
Yep, I like the use of `CMAKE_C_BYTE_ORDER`. I'm fine to take a dependency on cmake 3.20. Thanks for the PR!

BSzili:
I'm not at my usual setup so I tried to use WSL2 to test my changes. It turns out Ubuntu Bionic only has CMake 3.10.2 so I tried to use the older method. I added this to the root CMakeLists.txt, but it doesn't seem to change the generated Makefile:
```
include(TestBigEndian)
test_big_endian(IS_BIGENDIAN)
if (IS_BIGENDIAN)
    add_compile_definitions(BR_ENDIAN_BIG=1)
else()
    add_compile_definitions(BR_ENDIAN_LITTLE=1)
endif()
```
Maybe I put this in the wrong place?

madebr:
> Maybe I put this in the wrong place?

Because you're using `add_compile_options` (and not `target_compile_options(<target> PUBLIC <compile_optoins>)`),
you need to do the `add_compile_options` BEFORE any `add_library` calls.
Also, it needs be put AFTER the first call to `project`.
So somewhere in [this range](https://github.com/dethrace-labs/dethrace/blob/05f16d3066ce8eda39727cf10de0b06688d2f061/CMakeLists.txt#L16-L20).

> @madebr Sorry, I was so preoccupied with CMake that I forgot to answer your original question. I tested this on the Amiga.

Nice. Tested on a real Amiga or an emulated one?
Is OpenGL support done through software emulation?

BSzili:
> > Maybe I put this in the wrong place?
> 
> Because you're using `add_compile_options` (and not `target_compile_options(<target> PUBLIC <compile_optoins>)`), you need to do the `add_compile_options` BEFORE any `add_library` calls. Also, it needs be put AFTER the first call to `project`. So somewhere in [this range](https://github.com/dethrace-labs/dethrace/blob/05f16d3066ce8eda39727cf10de0b06688d2f061/CMakeLists.txt#L16-L20).
> 
I see, thanks. I'll try fiddle around with this after work. 

> > @madebr Sorry, I was so preoccupied with CMake that I forgot to answer your original question. I tested this on the Amiga.
> 
> Nice. Tested on a real Amiga or an emulated one? Is OpenGL support done through software emulation?

Only in emulation for now, the lack of software renderer is a big roadblock. I cobbled together some code to draw wireframe models, but it's quick just a test, very buggy and slow. I'm still contemplating what to do about this, as OpenGL is not an option on 25 year old hardware.

dethrace-labs:
that sounds awesome! Could you post a couple of screenshots!?

BSzili:
It's nothing to write home about, but here's one where it's not immediately obvious that I didn't do any clipping:

dethrace-labs:
looks awesome :)

dethrace-labs:
in the top-level `CMakelists.txt`, I added 
```include(TestBigEndian)
test_big_endian(IS_BIGENDIAN)``` above `find_package(SDL2 REQUIRED)`

Then in `BRSRC13/CMakeLists.txt, I added 
```
if(IS_BIGENDIAN)
    target_compile_definitions(brender PRIVATE BR_ENDIAN_BIG)
else()
    target_compile_definitions(brender PRIVATE BR_ENDIAN_LITTLE)
endif()
```

And then in `DETHRACE/CMakeLists.txt`:
```
if(IS_BIGENDIAN)
    target_compile_definitions(dethrace PRIVATE BR_ENDIAN_BIG)
else()
    target_compile_definitions(dethrace PRIVATE BR_ENDIAN_LITTLE)
endif()
```

It seems to set the value correctly as seen by the pre-processor. Maybe give that a try?

dethrace-labs:
May I post your screenshot on the @dethrace-labs twitter feed?

BSzili:
Of course, if you think it's interesting feel free to post it. I'll rebase the PR and implement the changes you have suggested.

dethrace-labs (approved):

PR #139, opened 2022-09-01
Little changes

madebr:
This is a collection of changes I made to DethRace while working on rec2 and transcoding some of this code to Python.
Functional difference is 0%.


dethrace-labs:
thanks! I've been toying with the idea of replacing BRSRC13 directory with a copy-paste of the closest "real" BRRender source then making our few minor tweaks to it to call out to harness. Could be a stepping stone to a fully-real BRRender backend

dethrace-labs (approved):

PR #140, opened 2022-09-01
Add support for fullscreen + never grab mouse + resize window

### LINKED ISSUE #72: Fullscreen
Please make a key to run in fullscreen mode.
-f --fullscreen

madebr:
- Switch to full screen (and back to windowed mode) when pressing <kbd>ALT</kbd>+<kbd>ENTER</kbd>
- Never grab the mouse (see discussion in discord)

The only issue I have with the current implementation is that it does not preserve aspect ratio.
I [tried the approach of this SO question](https://stackoverflow.com/questions/28587643/sdl2-how-to-keep-aspect-ratio-when-resizing-the-window), but the behavior is weird when going full screen and back. It also breaks when going full screen because it can resize the window to dimensions bigger then the screen size.

Instead, we probably just need to modify the gl renderer to draw its thing in the correct aspect ratio centrally and add black bars left and right. ==> This is what I ended up adding in this pr.

Fixes #72

### DIFF
--- a/src/harness/io_platforms/sdl_gl.c
+++ b/src/harness/io_platforms/sdl_gl.c
@@ -172,11 +172,6 @@ tRenderer* Window_Create(char* title, int width, int height, int pRender_width, 
         LOG_PANIC("Failed to create window");
     }
 
-    // Don't grab the mouse when a debugger is present
-    if (!OS_IsDebuggerPresent()) {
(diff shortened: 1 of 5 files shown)


dethrace-labs:
nice, will try it out over the weekend!

dethrace-labs (approved):

PR #145, opened 2022-09-07
Implement funkotronics

madebr:


dethrace-labs:
Nice! Possibly there is some sort of float to integer rounding that is different to the OG code? The movement seems to more jerky than the original? (for me at least)
  
original:
https://user-images.githubusercontent.com/78985374/189511846-0da38167-287b-424d-97b1-57e7be50fab3.mov

dethrace:
https://user-images.githubusercontent.com/78985374/189511896-d8c10f0d-c629-4152-8898-bc3524b586f1.mov

dethrace-labs (approved):
Going to merge now, and then deal with any jerkiness I still see as a separate issue 👍

PR #146, opened 2022-09-07
Implement edit functions + functions around camera + ...

madebr:
- Implement all flicplay functions (even those unused by the `.FLI` files in Carmageddon's game data)
- Implement edit functions for moving car
- Implement edit functions for adding actors/special volumes
- Implement depth related functions
- Fixes bug in car.c (introduced by me)


src/DETHRACE/common/depth.c
```diff
@@ -137,7 +148,11 @@ br_scalar EdgeU(br_angle pSky, br_angle pView, br_angle pPerfect) {
     br_scalar b;
     br_scalar c;
     LOG_TRACE("(%d, %d, %d)", pSky, pView, pPerfect);
-    NOT_IMPLEMENTED();
+
+    a = cosf(BrAngleToRadian(pPerfect));
+    b = sinf(BrAngleToRadian(pView));
+    c = cosf(BrAngleToRadian(pView));
+    return b * a * a / (BrAngleToRadian(pSky) * (1.f + c));
```
madebr, line 155: Please take a close look at this function as I'm not sure of it.

dethrace-labs: looks good to me

dethrace-labs:
I tried a couple of the accessory edit functions, I can't get them to work. Either locking on to the nearest or spawning a new accessory doesn't seem to work as expected (maybe they are being spawned in the wrong place?)

madebr:
> I tried a couple of the accessory edit functions, I can't get them to work. Either locking on to the nearest or spawning a new accessory doesn't seem to work as expected (maybe they are being spawned in the wrong place?)

Indeed, I missed to set the `translation` vector when the <kbd>SHIFT</kbd>` button was not pressed.
(The shift button is for axis aligning the actor when spawning on e.g. a hill, instead of aligning the object on the surface)

dethrace-labs (approved):

PR #147, opened 2022-09-07
Opponents aggresiveness

dethrace-labs:
- Implements credits and time bonus for damaging opponents
- Fixes opponent attack behavior so they actually chase the player now
- Implements cops


madebr:
Please merge #149 in this branch

PR #150, opened 2022-09-08
Fix address sanitizer error when loading "Industrial Action"

madebr:
The map "Industrial Action" references a pixelmap whose width is not a multiple of 4.
However, DethRace overrides the row stride to a multiple of 4 [here](https://github.com/dethrace-labs/dethrace/blob/f99be5f57a251f124a648fdca309460671ba93ab/src/DETHRACE/common/utility.c#L456) and [here](https://github.com/dethrace-labs/dethrace/blob/f99be5f57a251f124a648fdca309460671ba93ab/src/DETHRACE/common/utility.c#L472).
As a result of this, the GL renderer will do a buffer overread when [reading the pixels here](https://github.com/dethrace-labs/dethrace/blob/f99be5f57a251f124a648fdca309460671ba93ab/src/harness/renderers/gl/gl_renderer.c#L598-L604).

- To fix this, a helper function `align_pixelmap` is introduced that aligns each row.
- This pr also contains a patch for `resource.c` that extracts the "magic check" out of `UserToRes`.
  This is done to allow `BrResCheck` to not fail when run on a non-resource memory block.
  (I used this function while troubleshooting the Address Sanitizer error above)

### DIFF
--- a/src/DETHRACE/common/utility.c
+++ b/src/DETHRACE/common/utility.c
@@ -443,6 +443,30 @@ br_pixelmap* PurifiedPixelmap(br_pixelmap* pSrc) {
     NOT_IMPLEMENTED();
 }
 
+#if defined(DETHRACE_FIX_BUGS)
+static void align_pixelmap(br_pixelmap* pm) {
+    int i;
+    br_uint_16 original_row_bytes;
+    br_uint_16 new_row_bytes;
+    br_uint_8* original_pixels;
+    br_uint_8* new_pixels;
+
+    original_row_bytes = pm->row_bytes;
+    if (original_row_bytes % sizeof(int) == 0 || pm->height <= 1) {
+        return;
+    }
+    new_row_bytes = (original_row_bytes + sizeof(int) - 1) & ~(sizeof(int) - 1);
+    original_pixels = pm->pixels;
+    new_pixels = BrResAllocate(pm, new_row_bytes * pm->height, BR_MEMORY_PIXELS);
+    for (i = 0; i < pm->height; i++) {
+        memcpy(new_pixels + i * new_row_bytes, original_pixels + i * original_row_bytes, original_row_bytes);
+    }
+    BrResFree(original_pixels);
+    pm->pixels = new_pixels;
+    pm->row_bytes = new_row_bytes;
+}
+#endif
+
 // IDA: br_pixelmap* __usercall DRPixelmapLoad@<EAX>(char *pFile_name@<EAX>)
 br_pixelmap* DRPixelmapLoad(char* pFile_name) {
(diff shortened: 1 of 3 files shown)


src/DETHRACE/common/utility.c
```diff
@@ -453,7 +477,11 @@ br_pixelmap* DRPixelmapLoad(char* pFile_name) {
     if (the_map != NULL) {
         the_map->origin_x = 0;
         the_map->origin_y = 0;
+#if defined(DETHRACE_FIX_BUGS)
```
dethrace-labs, line 480: instead of introducing `align_pixels`, what if we just skipped overriding the row_bytes?

```
#if !defined(DETHRACE_FIX_BUGS)
the_map->row_bytes = (the_map->row_bytes + sizeof(int) - 1) & ~(sizeof(int) - 1);
#endif
```

src/DETHRACE/common/utility.c
```diff
@@ -443,6 +443,30 @@ br_pixelmap* PurifiedPixelmap(br_pixelmap* pSrc) {
...
+    br_uint_16 original_row_bytes;
+    br_uint_16 new_row_bytes;
+    br_uint_8* original_pixels;
+    br_uint_8* new_pixels;
+
+    original_row_bytes = pm->row_bytes;
+    if (original_row_bytes % sizeof(int) == 0 || pm->height <= 1) {
+        return;
+    }
+    new_row_bytes = (original_row_bytes + sizeof(int) - 1) & ~(sizeof(int) - 1);
```
dethrace-labs, line 458: i wonder if everywhere we are using `sizeof(int) - 1) & ~(sizeof(int) - 1)` we should be using `int32_t` instead of `int`. We always expect that value to be "4" regardless of the size of an int

madebr: Yes good point.

src/DETHRACE/common/utility.c
madebr: In general, the idea of aligning data is to accelerate access times.
I remember reading some platforms not supporting unaligned access, requiring the operating system to handle these.

src/DETHRACE/common/utility.c
madebr: Added this in b9d600b

src/DETHRACE/common/utility.c
madebr: I applied your suggestion in b9d600b

Looking back at my initial custom `allign_pixelmap` function, I think I also needed to modify the `br_pixelmap`'s `flags` member as follows:

```c
pm->flags &= ~BR_PMF_LINEAR;

dethrace-labs (approved):

PR #152, opened 2022-09-09
Implement DoEndSummary2 (aka Damage Gallery)

madebr:
This pr implements the post-race "Damage Gallery".

Screenshot from 2022-09-09 15-32-46
Screenshot from 2022-09-09 15-32-01

### DIFF
--- a/src/DETHRACE/common/racesumm.c
+++ b/src/DETHRACE/common/racesumm.c
@@ -2,17 +2,23 @@
 #include "brender/brender.h"
 #include "crush.h"
 #include "cutscene.h"
+#include "displays.h"
 #include "flicplay.h"
 #include "globvars.h"
+#include "globvrkm.h"
 #include "globvrpb.h"
 #include "grafdata.h"
 #include "graphics.h"
 #include "harness/config.h"
 #include "harness/trace.h"
+#include "input.h"
 #include "intrface.h"
 #include "loading.h"
+#include "main.h"
 #include "network.h"
+#include "opponent.h"
 #include "pd/sys.h"
+#include "raycast.h"
 #include "s3/s3.h"
 #include "sound.h"
 #include "utility.h"
@@ -473,7 +479,17 @@ void PrepareBoundingRadius__racesumm(br_model* model) {
     int v;
     br_vertex* vp;
     LOG_TRACE("(%p)", model);
-    NOT_IMPLEMENTED();
+
+    max = 0.f;
+    for (v = 0; v < model->nvertices; v++) {
+        vp = &model->vertices[v];
+        d = BrVector3LengthSquared(&vp->p);
+        if (d > max) {
+            max = d;
+        }
+    }
+    d = sqrtf(max);
+    model->radius = d;
 }
 
 // IDA: void __cdecl BuildWrecks()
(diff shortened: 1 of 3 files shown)


src/DETHRACE/common/racesumm.c
```diff
@@ -564,21 +772,153 @@ void DamageScrnDraw(int pCurrent_choice, int pCurrent_mode) {
...
     char* name;
     LOG_TRACE("(%d, %d)", pCurrent_choice, pCurrent_mode);
-    NOT_IMPLEMENTED();
+
+    if (((pCurrent_choice == 0 && pCurrent_mode == 0) || !gDone_initial) && (gWreck_zoomed_in < 0 && gWreck_selected >= 0)) {
+        sel_actor = BrActorAllocate(BR_ACTOR_MODEL, NULL);
+        sel_actor->render_style = BR_RSTYLE_BOUNDING_EDGES;
+        sel_actor->render_style = BR_RSTYLE_NONE;  // FIXME: remove this line once BR_RSTYLE_BOUNDING_EDGES rener style has been implemente
+        sel_actor->model = gWreck_array[gWreck_selected].actor->model;
+         BrActorAdd(gWreck_array[gWreck_selected].actor, sel_actor);
```
madebr, line 781: The line:
```c
sel_actor->render_style = BR_RSTYLE_NONE;
```
is the only difference with Carmageddon, because rendering bounding boxes is not implemented (yet) by the opengl renderern.

dethrace-labs:
I'm getting some compiler warnings on this new code - 

few of these ones:
```
src/DETHRACE/common/racesumm.c:1005:34: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
                    difference = abs(gWreck_array[i].pos_x - gWreck_array[gWreck_selected].pos_x);
                                 ^
```

```
src/DETHRACE/common/racesumm.c:1046:9: warning: variable 'old_mouse_x' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
    if (gWreck_zoomed_in < 0) {
```

madebr:
> I'm getting some compiler warnings on this new code -

We really need to reduce the number of warnings to 0.
Because right now potential critical warnings are drown in silly warnings.
Right now, CI is just there to signal fail/no fail.
For rec2, I make it a point to have 0 warnings so every warning gets my full attention.

> 
> ```
> src/DETHRACE/common/racesumm.c:1005:34: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
>                     difference = abs(gWreck_array[i].pos_x - gWreck_array[gWreck_selected].pos_x);
>                                  ^
> ```


This one is intentional, to silence the warning an explicit `(int)` needs to be added.

> 
> ```
> src/DETHRACE/common/racesumm.c:1046:9: warning: variable 'old_mouse_x' is used uninitialized whenever 'if' condition is false [-Wsometimes-uninitialized]
>     if (gWreck_zoomed_in < 0) {
> ```
> 
> for things like `old_mouse_x`, in the past I've just initialized them to `0` in their declaration to get a clean build

This needs something like that. I don't think there is a general way of tagging variables to say they are written by a function. Microsoft has [SAL annotations](https://docs.microsoft.com/en-us/previous-versions/visualstudio/visual-studio-2010/ms235402(v=vs.100)?redirectedfrom=MSDN) but idk about other compilers.

dethrace-labs:
> We really need to reduce the number of warnings to 0.
Agree. We used to have this, but over time the number has crept up. Its not bad on gcc/clang, but I think on windows (where I don't spend much time) is worse.

Need to be more intentional with the warnings we enable probably, things like "variable initialized but not used" shouldn't be flagged - we are replicating existing logic rather than writing everything as cleanly as possible.

dethrace-labs (approved):

PR #160, opened 2022-09-15
damage_magnitude_accumulator is only available when a car is oppo, net_human or local human

madebr:
This often caused the address sanitizer to decide for me the race is over :smile:


dethrace-labs (approved):
good find, thanks!

PR #162, opened 2022-09-15
Use nearest_section in FindNearestGeneralSection

madebr:
After the for-loop preceding this array access, `section_no` was always set to the `section_no`.
Hence this would always access out-of-bounds.

Found by the trusty address sanitizer.


dethrace-labs:
Doh, another good find! I'm not sure what happened here - my view of the OG code is using `nearest_section` correctly, I must have mistyped it somehow :(

dethrace-labs (approved):

madebr:
Bugs happen, otherwise we would be all out of jobs :smile:

PR #169, opened 2022-09-16
No more gcc/clang warnings

madebr:
This pr fixes all gcc/clang warnings.
`-Werror` for Apple/clang is not (yet) enabled, as I would like to fix MSVC warnings in a future pr first.


madebr:
I also fixed a few `-Wextra` warnings (not enabled yet).
99% of the remaining warnings are due to `-Wsign-compare`.

src/DETHRACE/common/opponent.c
```diff
@@ -1787,33 +1794,32 @@ int MassageOpponentPosition(tOpponent_spec* pOpponent_spec, int pMassage_count)
...
         direction_v.v[2] = -pOpponent_spec->car_spec->car_master_actor->t.t.mat.m[2][2];
         if (pMassage_count % 4 >= 2) {
-
             BrVector3Cross(&displacement, &positive_y_vector, &direction_v);
             BrVector3Normalise(&displacement, &displacement);
             BrVector3Scale(&displacement, &displacement, (pMassage_count / 4) * 0.1f);
         } else {
-            BrVector3Normalise(&displacement, &displacement);
-            BrVector3Scale(&displacement, &displacement, (pMassage_count / 4) * 0.5f);
+            BrVector3Normalise(&direction_v, &direction_v);
```
dethrace-labs, line 1814: I think you're right - the original code here is not quite right. I think it should be this though (slightly different from what you propose:

```
BrVector3Normalise(&displacement, &direction_v);
BrVector3Scale(&displacement, &displacement, (pMassage_count / 4) * 0.5f);
```

src/DETHRACE/common/opponent.c
```diff
@@ -1787,33 +1794,32 @@ int MassageOpponentPosition(tOpponent_spec* pOpponent_spec, int pMassage_count)
...
+            BrVector3Scale(&displacement, &direction_v, (pMassage_count / 4) * 0.5f);
```
madebr, line 1815: My idea was that a direction is almost always a unit vector, which is then scaled.
But the code is 100% equivalent since `direction_v` is not used later.
```suggestion
            BrVector3Normalise(&displacement, &direction_v);
            BrVector3Scale(&displacement, &displacement, (pMassage_count / 4) * 0.5f);
```

dethrace-labs: I thought exactly the same thing :) My initial thought was that `displacment` and `direction_v` were just around the wrong way across the entire function, but later on `displacement` is used to update the `car_trans`, so I _think_ we have it around the right way!

What you changed it to in 1de95b0 matches my reading of the OG code

dethrace-labs (approved):

PR #175, opened 2022-09-22
Fix buffer overload when loading more then 100 pixelmaps on 64-bit system

madebr:
This fixes a `Bad resource header from user` error I received when testing DethRace with [The MELD PACK](https://rr2000.cwaboard.co.uk/carmageddon-1).

Running with the address sanitizer, I was able to pin point the error immediately.


dethrace-labs:
nice catch!

dethrace-labs (approved):

PR #205, opened 2022-10-02
Workflow Updates

dethrace-labs:
- simplifies workflow yaml
- stops double-running workflow on pull request and push. Workflow can be forced to run without opening a PR by naming your branch in the format `ci-*`
- generates downloadable artifacts on each push to `main`
- releases named in the form of _Dethrace 0.4.0_ rather than _Dethrace v0.4.0_ to match the more common convention


.github/scripts/build-macos.sh
```diff
@@ -9,4 +15,6 @@ cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTS=ON -DDETHRACE_WERROR=ON -B
 cmake --build build -- -j 4
```
madebr, line 15: I think you need to add `-DCMAKE_OSX_ARCHITECTURES=x86_64` (or `amd64` idk) to be sure this does not build an armv8 executable.
Speaking of armv8, perhaps add `armv8` to the build table for macos?

See [`CMAKE_OSX_ARCHITECTURES`](https://cmake.org/cmake/help/latest/variable/CMAKE_OSX_ARCHITECTURES.html).
I think you can build a "fat" executable by doing `-DCMAKE_OSX_ARCHITECTURES="x86_64;armv8"`, but don't quote me on that :)

.github/scripts/build-macOS.sh
```diff
@@ -0,0 +1,20 @@
...
+brew install SDL2
+
+# build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTS=ON -DDETHRACE_WERROR=ON -B build
+cmake --build build -- -j 4
+
+# package artifact
+tar -czvf dethrace-${BUILD_TAG}-darwin-amd64.tar.gz build/dethrace
+
+echo "::set-output name=filename::dethrace-${BUILD_TAG}-darwin-amd64.tar.gz"
```
madebr, line 20: This is just cosmetic, but can you please add a final new line to all files to avoid GitHub adding this marker.

dethrace-labs: thanks. I moved to a new laptop and forgot to turn on the auto-insert newlines setting 👍

madebr: On my dev machine, I added a `.editorconfig` in the root folder of my user home dir containing:
```
root = true

[*]
end_of_line = lf
insert_final_newline = true

[Makefile]
indent_style = tab
indent_size = 4

[*.py]
indent_style = space
tab_width = 4
```
And all my ide's are configured to use editorconfig files.

When some project has another default, I add an `.editorconfig` file in that projects root directory.
To avoid git picking up that file, I added `.editorconfig` to `.gitignore` in my user home directory.

.github/scripts/build-macOS.sh
```diff
@@ -0,0 +1,20 @@
...
+set -e
+
+if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
+    BUILD_TAG=${GITHUB_REF_NAME}
+else 
+    BUILD_TAG=$(git rev-parse --short HEAD)
+fi
+
+# install deps
+brew install SDL2
```
madebr, line 11: On Linux and Macos, we implicitly assume people have SDL2 somehow installed.
Should we add `libSDL2.(dylib|so)` to the archive?
In that case, we might need to fix rpaths.

Food for thoughts, and for a later pr.

.github/workflows/workflow.yaml
```diff
@@ -1,76 +1,42 @@
...
         with:
           version: 1.10.2
       - name: Build
-        env:
-          CC: ${{ steps.vars.outputs.cc }}
-          CXX: ${{ steps.vars.outputs.cxx }}
-          MATRIX_PLATFORM: ${{ matrix.platform }}
-        run: .github/scripts/build-linux.sh
+        id: build
+        run: .github/scripts/build-${{ runner.os }}.sh
```
madebr, line 26: Can we collapse the windows job into this job by adding the build script into the matrix?
e.g.:
```yml
matrix:
    include:
      - os: 'ubuntu-latest'
        script: 'build-linux.sh'
      - os: 'macos-latest'
        script: 'build-macos.sh'
      - os: 'windows-latest'
        script: 'build-msvc.ps1'
```

.github/scripts/build-macos.sh
dethrace-labs: thanks. github actions don't yet support arm osx builds (https://github.com/actions/runner-images/issues/2187), so I've set it just to `x86_x64`

.github/scripts/build-macOS.sh
dethrace-labs: yep - I ran into this problem on a different laptop where libsdl2 was not in the expected place. Definitely an issue to fix - I think as part of generating a fat `.app` package for macos

.github/scripts/build-macos.sh
madebr: We don't need M1 hardware. We just want to cross build to armv8.

.github/workflows/workflow.yaml
dethrace-labs: I considered that, but for windows we currently build with an additional platform matrix, so I wasn't sure how we should deal with that. Probably the easiest way would be to drop support for 32 bit windows builds, but its quite nice being able to build for the original architecture

madebr: Let's leave 32-bit builds enabled, if not for checking our pointer arithmetic is correct.

You mean the `ilammy/msvc-dev-cmd@v1.4.1` step? That can be made conditional on the os using
```yml
      - uses: ilammy/msvc-dev-cmd@v1.4.1
        if: contains(matrix.os, 'windows')
        with:
          arch: ${{ matrix.platform }}
```

madebr: And the platforms can then only be added to the Windows items.

```yml
    include:
      - os: 'ubuntu-latest'
        script: '.github/scripts/build-linux.sh'
      - os: 'macos-latest'
        script: '.github/scripts/build-macos.sh'
      - os: 'windows-latest'
        script: '.github\\scripts\\build-msvc.ps1'
        platform: 'x86'
      - os: 'windows-latest'
        script: '.github\\scripts\\build-msvc.ps1'
        platform: 'amd64'
```

dethrace-labs: no, I mean for windows we build x86 and x64. Ah, just saw your other comment, yep that should owrk

.github/scripts/build-macos.sh
madebr: Actually, we do when we run the tests. So you're right.

.github/workflows/workflow.yaml
```diff
@@ -1,177 +1,78 @@
...
-        with:
-          name: artifacts
-          path: darwin-amd64.tar.gz
-
-  build-windows:
-    runs-on: windows-2019
-    strategy:
-      matrix:
-        platform: [x86, amd64]
+        os: [ubuntu-latest, macos-latest, windows-latest]
```
madebr, line 17: This should be removed

madebr:
```suggestion
```

.github/scripts/build-macos.sh
```diff
@@ -1,12 +1,20 @@
...
 # build
-cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DBUILD_TESTS=ON -DDETHRACE_WERROR=ON -B build
+cmake -DCMAKE_BUILD_TYPE=RelWithDebInfo -DCMAKE_OSX_ARCHITECTURES=x86_64 -DBUILD_TESTS=ON -DDETHRACE_WERROR=ON -B build
 cmake --build build -- -j 4
 
 # package artifact
-tar -czvf darwin-amd64.tar.gz build/dethrace
+tar -czvf dethrace-${BUILD_TAG}-darwin-amd64.tar.gz build/dethrace
+
+echo "::set-output name=filename::dethrace-${BUILD_TAG}-darwin-amd64.tar.gz"
```
madebr, line 20: Something what bothers me with the current Linux release is that the tarball contains a `build` folder.
How about we create a `dethrace-${BUILD_TAG}-darwin-amd64` directory, and tar that directory?

```suggestion
releasename="dethrace-${BUILD_TAG}-darwin-amd64"

rm -rf "$releasename"
mkdir "$releasename"
cp build/dethrace "$releasename/dethrace"
tar -czvf "$releasename.tar.gz" "$releasename"

echo "::set-output name=filename::"$releasename.tar.gz"
```

madebr: The same applies to the Macos release binary.
The Windows zip contains no directories.

.github/workflows/workflow.yaml
```diff
@@ -1,177 +1,82 @@
...
+            script: build-linux.sh
+          - os: macos-latest
+            script: build-macos.sh
+          - os: windows-latest
+            script: build-msvc.ps1
+        exclude:
+          - os: ubuntu-latest
+            platform: x86
+          - os: macos-latest
+            platform: x86
```
madebr, line 32: Doesn't this work?
```yml
      matrix:
        include:
          - os: ubuntu-latest
            script: build-linux.sh
          - os: macos-latest
            script: build-macos.sh
          - os: windows-latest
            script: build-msvc.ps1
            platform: amd64
          - os: windows-latest
            script: build-msvc.ps1
            platform: x86
```

In the SDL project, [we don't use `include` but `platform`](https://github.com/libsdl-org/SDL/blob/19ecb64e0dfea5616798f77d8262715f0d722038/.github/workflows/main.yml#L17-L28).

dethrace-labs: thanks for the platform link, I didn't know you could define whole objects as matrix items. The include idea didn't seem to work, but this did.

PR #213, opened 2022-10-16
Tidy up debug logging

dethrace-labs:
- disable diagnostics by default (`--enable-diagnostics` command line parameter to enable)
- print OpenGL information strings 
- remove random debug log lines that are not longer useful


src/DETHRACE/common/loading.c
```diff
@@ -3295,7 +3295,7 @@ int GetCDPathFromPathsTxtFile(char* pPath_name) {
 int TestForOriginalCarmaCDinDrive() {
     LOG_TRACE("()");
 
-    // Jeff: the symbol dump didn't include any local variable information.
+    // The symbol dump didn't include any local variable information.
```
madebr, line 3298: For C90 compliance, these variables should be declared before the `LOG_TRACE`.

madebr (approved):

PR #217, opened 2022-10-18
Original "high resolution" support

### LINKED ISSUE #159: "-hires" flag is not implemented
Some code paths that are required for high res mode need to be implemented:
- pixel doubling (?)
- gl renderer assumptions
- other unknown things

dethrace-labs:
Implements support for `-hires` command line arg rendering at 640x480 and using more detailed pixmaps from the `64x48x8` directory.

Fixes #159 

Screen Shot 2022-10-19 at 9 47 02 AM

### DIFF
--- a/src/BRSRC13/CORE/DOSIO/dosio.c
+++ b/src/BRSRC13/CORE/DOSIO/dosio.c
@@ -7,6 +7,5 @@
 br_pixelmap* DOSGfxBegin(char* setup_string) {
     LOG_TRACE("(\"%s\")", setup_string);
 
-    Harness_Hook_DOSGfxBegin();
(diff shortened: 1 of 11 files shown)


madebr:
Looks good! Only issue I can see is the mouse pointer not scaled correctly.

dethrace-labs:
I don't normally use the mouse, so I didn't notice. I did some digging and the OG has the same behavior!

madebr:
You're right. The mouse pointer is hidden, so I didn't notice.

Let's fix this with `DETHRACE_FIX_BUGS`. #218 tracks this.

madebr (approved):

mrt123:
how would you supply -hires command line arg for MacOS version?

Issue #187, opened 2022-09-29
Missing texture on Mayan Mayhem track


rafalp:
<img width="1440" alt="Zrzut ekranu 2022-09-29 o 18 57 22" src="https://user-images.githubusercontent.com/750553/193093489-7e695c72-bd5c-4849-aebc-d544d4150e34.png">

<img width="1440" alt="Zrzut ekranu 2022-09-29 o 18 59 37" src="https://user-images.githubusercontent.com/750553/193093531-28aa98b6-1041-4ab5-9cee-8ce3f3128579.png">

dethrace-labs:
Can reproduce, looking into it!

PR #222, opened 2022-10-26
Re-enables back face culling

dethrace-labs:
The cause of #187 is weird track geometry where two back-facing triangles with no texture are laid on top of two front-facing textured triangles. The OG game culls the back-facing triangle, so it is not visible in game. (Ref: https://twitter.com/toshiba_3/status/1584152161386389504?s=20&t=g1bQ8KAAtKvn2zy1Qw6PUw)

image

As we are not culling, we render the back facing triangles, which causes the appearance of a missing texture in that part of the track!

We originally removed culling because it caused sides of streetlight poles to be invisible when viewed from the rear. Since then, we fixed/implemented `RemoveDoubleSided` which causes every face with a `BR_MATF_TWO_SIDED` material to be duplicated and faced backwards. We can now cull back faces without any apparent graphical issues.

Fixes: #187

### DIFF
--- a/src/harness/renderers/gl/gl_renderer.c
+++ b/src/harness/renderers/gl/gl_renderer.c
@@ -215,13 +215,13 @@ void GLRenderer_Init(int width, int height, int pRender_width, int pRender_heigh
     LoadShaders();
     SetupFullScreenRectGeometry();
 
-    // opengl config
+    // config
     glDisable(GL_BLEND);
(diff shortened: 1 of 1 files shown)


madebr (approved):
Hurray! Less triangles to draw!

dethrace-labs:
Yep, its win-win 🎉

PR #228, opened 2022-10-30
Fix sounds

dethrace-labs:
- Fixes sounds should be stopped when the emitter is too far way (weren't calling `S3UpdateSoundSource` inside `S3Service`)
- Fixes DirectSound decibels to linear conversion by using `ma_volume_db_to_linear` 
- Adds `--volume-multiplier` command line arg to set overall volume gain. (windows seems to match OG, on OSX the audio seems too loud at the default gain of 1.0


src/S3/3d.c
```diff
@@ -517,21 +516,21 @@ int S3Calculate3D(tS3_channel* chan, int pIs_ambient) {
...
         if (v11 <= 2.0f) {
-            if (v11 < 0.5) {
-                v11 = 0.5;
+            if (v11 < 0.5f) {
+                v11 = 0.5f;
             }
         } else {
-            v11 = 2.0;
+            v11 = 2.0f;
         }
```
madebr, line 526: While you're looking at this function:
```suggestion
        if (v11 > 2.0f) {
            v11 = 2.0f;
        else if (v11 < 0.5f) {
            v11 = 0.5;
        }
```

dethrace-labs: yep 👍

Issue #220, opened 2022-10-20
Skidmarks from wall driving are horizontal


rafalp:
If you drive on the wall, skidmarks will be rendered horizontally vs vertically:

<img width="1440" alt="Zrzut ekranu 2022-10-20 o 00 15 05" src="https://user-images.githubusercontent.com/750553/197062493-31903d72-44af-4a12-b403-02acf2fad9c2.png">

madebr:
This is fixed by #233

PR #233, opened 2022-11-06
Fix skidmark orientation

zear:
Don't reset skidmark normal to the horizontal plane, but rather use the one received from the wheel.

This fixes incorrect rendering of skidmarks on sloped surfaces (#220).

Before:

https://user-images.githubusercontent.com/510643/200192096-4ade21ad-a683-42af-9394-59016e16c75f.mp4

After:

https://user-images.githubusercontent.com/510643/200192095-d70e2e9b-5793-4a08-96d4-caac67d1d604.mp4


madebr (approved):
100% correct.
These assignments are not available in the OG game.
They are probably left-over debug code.

Thanks!

dethrace-labs (approved):
Oops! Yes, I left them in by mistake when debugging the initial implementation! Good catch :)

PR #235, opened 2022-11-08
Show Squad Car in the wreck gallery

### LINKED ISSUE #208: Cops car is not rendered on damage gallery
Finish race with cop car and go to damage gallery, it won't be displayed:

Zrzut ekranu 2022-10-10 o 00 04 51
Zrzut ekranu 2022-10-10 o 00 04 53

zear:
Fix OG bug, where APC car would have its vertices and faces removed, preventing it from being rendered in the wreck gallery. Fixes #208


dethrace-labs (approved):
nice fix! I never noticed this bug in OG. I guess they wanted to keep memory usage down as much as possible by throwing away the original vertex data

PR #237, opened 2022-11-09
Fixes + add TestAutoSpecialVolume (=underwater)

### LINKED ISSUE #194: No difference in handling underwater
Underwater ability doesn't do anything at the moment as cars handle the same underwater as they do on land.

### LINKED ISSUE #218: Cursor position scaled incorrectly in `-hires` mode
This behavior also occurs in the original game.

https://user-images.githubusercontent.com/78985374/196538847-c9dec66c-3583-413c-af8c-33a9c56905fb.mov

madebr:
- Fix crash at the end of a race when playing the splatpack (non-xmas) demo: the splatpack demo plays a smk, whereas the xmas demo displays a flic
- Simplify a few functions by switching if/else branches to reduce indentation and using `BrXXX` functions/macro's
- Fix `DecodeLine2` + `tools/decode_datatxt.py` so the coments become readable
- Implement `StopGroovidelic`
- Implement `TestAutoSpecialVolume: this increases the drag underwater
    Fixes #194 
- I removed the stub in `DrawTheGlow`.
   This change might be a bit controversial as it might introduces more crashes.
   But I think this is good so we can find the bug sooner.
   Either case, the crash in `SmokeLine` happens on my system too with stubbing this function.
- Fix address sanitizer warning when riding the APC car
- add `--full-screen` command line option to start DethRace in full screen mode
- Make the mouse pointer behave well when using hires by scaling the coordinates
    Fixes #218
- Fixes `IWANTTOFIDDLE`. Tested with Carmageddon 1 full game data by encoding/decoding the data a few times.

### DIFF
--- a/src/DETHRACE/common/loading.c
+++ b/src/DETHRACE/common/loading.c
@@ -1687,7 +1687,7 @@ void MungeWindscreen(br_model* pModel) {
 void SetModelFlags(br_model* pModel, int pOwner) {
     LOG_TRACE("(%p, %d)", pModel, pOwner);
 
-    if (pModel && pModel->nfaces) {
+    if (pModel != NULL&& pModel->nfaces != 0) {
 #if defined(DETHRACE_FIX_BUGS) /* Show Squad Car in the wreck gallery. */
         if (gAusterity_mode) {
 #else
--- a/src/DETHRACE/common/loading.c
+++ b/src/DETHRACE/common/loading.c
@@ -1796,6 +1796,7 @@ void LoadCar(char* pCar_name, tDriver pDriver, tCar_spec* pCar_spec, int pOwner,
     }
     if (*pDriver_name != '\0') {
 #if defined(DETHRACE_FIX_BUGS)
+        // Make sure to not read and write out of bounds.
         memcpy(pCar_spec->driver_name, pDriver_name, MIN(sizeof(pCar_spec->driver_name), strlen(pDriver_name)));
 #else
         memcpy(pCar_spec->driver_name, pDriver_name, sizeof(pCar_spec->driver_name));
--- a/src/DETHRACE/common/loading.c
+++ b/src/DETHRACE/common/loading.c
@@ -2430,11 +2430,11 @@ void LoadRaces(tRace_list_spec* pRace_list, int* pCount, int pRace_type_index) {
     *pCount = number_of_racers;
     fclose(f);
     j = 0;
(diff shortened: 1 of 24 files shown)


src/DETHRACE/common/loading.c
```diff
@@ -1795,7 +1795,11 @@ void LoadCar(char* pCar_name, tDriver pDriver, tCar_spec* pCar_spec, int pOwner,
         FatalError(kFatalError_FileCorrupt_S, pCar_name);
     }
     if (*pDriver_name != '\0') {
+#if defined(DETHRACE_FIX_BUGS)
```
dethrace-labs, line 1798: could you add a comment documenting what the bug is that we are fixing?

src/DETHRACE/common/utility.c
```diff
@@ -144,45 +142,52 @@ void EncodeLine(char* pS) {
...
-        pS[len] = 0;
+        len--;
+        pS[len] = '\0';
     }
 
     seed = len % 16;
+
     for (i = 0; i < len; i++) {
         c = pS[i];
+#if defined(DETHRACE_FIX_BUGS)
```
dethrace-labs, line 153: add a comment documenting the original bug behavior?

src/DETHRACE/common/loading.c
madebr: The OG game (carm95.exe at least) does not do this.
It keeps the original key and does not decode the comments with the alternative key.
I'll add this as a comment.

src/DETHRACE/common/utility.c
madebr: The OG game (carm95.exe at least) does not do this.
It keeps the original key and does not decode the comments with the alternative key.
I'll add this as a comment.

dethrace-labs: that would be great, thanks.

src/DETHRACE/common/utility.c
```diff
@@ -144,45 +142,52 @@ void EncodeLine(char* pS) {
...
         if (i >= 2) {
             if (pS[i - 1] == '/' && pS[i - 2] == '/') {
                 key = (char*)gOther_long_key;
             }
         }
+#endif
         if (gEncryption_method == 1) {
             if (c == '\t') {
-                c = 0x80;
+                c = 0x9f;
```
dethrace-labs, line 162: hmm, good spotting. Dont know how I got encryption methods round the wrong way! 🤔

dethrace-labs (approved):
LGTM 👍

PR #238, opened 2022-11-10
Embed GLSL files during build

dethrace-labs:
- Adds separate files for our GLSL shaders
- Adds generation step for each one to convert it into a header file during the cmake pipeline
- Begin to separate out a section at the top of the logs that we will ask users to copy when filing issues


src/harness/CMakeLists.txt
```diff
@@ -61,14 +61,26 @@ target_sources(harness PRIVATE
 )
 
 if (IO_PLATFORM STREQUAL "SDL_OpenGL")
+    # generate embedded glsl resources
+    foreach(elem resources/framebuffer_vert.glsl resources/framebuffer_frag.glsl resources/3d_vert.glsl resources/3d_frag.glsl)
+        add_custom_command(
+            OUTPUT "${CMAKE_CURRENT_BINARY_DIR}/${elem}.h"
+            COMMAND ${CMAKE_COMMAND} -DSOURCE_DIR=${CMAKE_CURRENT_SOURCE_DIR} -DFILE="${elem}" -P ${CMAKE_SOURCE_DIR}/cmake/EmbedResource.cmake
```
madebr, line 68: `CMAKE_SOURCE_DIR` ponit to the root of a build tree, where we want the root of the the dethrace project.

```suggestion
            COMMAND "${CMAKE_COMMAND}" "-DSOURCE_DIR=${dethrace_SOURCE_DIR}" "-DFILE=${elem}" -P "${dethrace_SOURCE_DIR}/cmake/EmbedResource.cmake"
```

src/harness/resources/3d_vert.glsl
```diff
@@ -0,0 +1,20 @@
...
+uniform mat4 u_model;
+uniform mat4 u_view;
+uniform mat4 u_projection;
+
+void main() {
+    v_frag_pos = vec3(u_model * vec4(a_pos, 1.0));
+    v_normal = a_normal;
+    v_tex_coord = a_uv;
+    gl_Position = u_projection * u_view * vec4(v_frag_pos, 1.0);
+}
```
madebr, line 20: Missing final newline

src/harness/CMakeLists.txt
dethrace-labs: I understand the `"`, but why `dethrace_SOURCE_DIR`? These are resources specific to `harness`, and should only be available to `#include` from harness code?

madebr: Looking at it, use of `CMAKE_CURRENT_SOURCE_DIR` is correct.
But `CMAKE_SOURCE_DIR` isn't: it should be either `PROJECT_SOURCE_DIR` or `dethrace_SOURCE_DIR`.

dethrace-labs: are you suggesting to move `cmake/EmbedResource.cmake` to `src/DETHRACE/cmake/EmbedResource.cmake`?

madebr: No, if you happen to include dethrace as a subproject for your own project, `CMAKE_SOURCE_DIR` points to the root source directory. Not to the root directory of dethrace.

dethrace-labs: ah, yep, I got it I think.

madebr (approved):
Works on my pc :)

PR #240, opened 2022-11-11
Fix oil spill rendering

zear:
Due to the wrong order of vertices in the model, oil spill object is subject to back face culling, preventing it from being rendered. Make the oil spill front faced w/ regards to the floor.

Note: Setting `BR_MATF_TWO_SIDED` for the material flags does not solve this problem - the oil spill is still invisible. Has the original order of vertices been taken from OG disassembly?

### DIFF
--- a/src/DETHRACE/common/oil.c
+++ b/src/DETHRACE/common/oil.c
@@ -41,6 +41,8 @@ void InitOilSpills() {
         the_material->flags |= BR_MATF_LIGHT;
         the_material->flags |= BR_MATF_PERSPECTIVE;
         the_material->flags |= BR_MATF_SMOOTH;
+	// TODO: added by dethrace, investigate why oil spills in OG do not need this flag set to render correctly
+        the_material->flags |= BR_MATF_TWO_SIDED;
         the_material->index_range = 0;
         the_material->colour_map = NULL;
         BrMatrix23Identity(&the_material->map_transform);
--- a/src/harness/renderers/gl/gl_renderer.c
+++ b/src/harness/renderers/gl/gl_renderer.c
@@ -500,6 +500,12 @@ void setActiveMaterial(tStored_material* material) {
         glUniform1i(uniforms_3d.light_value, -1);
     }
 
+    if (material->flags & (BR_MATF_TWO_SIDED | BR_MATF_ALWAYS_VISIBLE)) {
+        glDisable(GL_CULL_FACE);
+    } else {
+        glEnable(GL_CULL_FACE);
+    }
+
     if (material->pixelmap) {
         tStored_pixelmap* stored_px = material->pixelmap->stored;
         if (stored_px != NULL) {


madebr:
I just checked and the ordering is 1:1 as used in `CARM95.EXE`.

I think the issue is our renderer does not respect the `BR_MATF_TWO_SIDED` flag.
Grep'ing for `BR_MATF_TWO_SIDED` gives no matches in neither `src/BRSRC13` and `src/harness`.

madebr:
Btw, you're always welcome to join the Carmageddon DIscord server to discuss various problems.

madebr:
I also just noticed the non-rendering of an oil spill, even though it caused the tyres to slip and an oil track to form.
I'm wondering what is the OG behavior.

zear:
Updated the patch with gl renderer support of `BR_MATF_TWO_SIDED`.
This works by disabling culling for the time the two-sided model is being rendered, and then re-enabling it afterwards. I am not familiar with OpenGL API, so hopefully this is the right approach. Seems to work just fine with my renderer.

The oil spills now also render at the same time other back-faced geometry is properly culled (#187):
spill


> Btw, you're always welcome to join the Carmageddon Discord server to discuss various problems.

Thanks, but I'm not using walled garden ecosystems. Unless you guys have an IRC channel :)

src/harness/renderers/gl/gl_renderer.c
```diff
@@ -549,6 +550,10 @@ void GLRenderer_Model(br_actor* actor, br_model* model, br_matrix34 model_matrix
         glUniform1i(uniforms_3d.light_value, -1);
     }
 
+    twosided = actor->material && actor->material->flags & BR_MATF_TWO_SIDED;
+    if (twosided)
+        glDisable(GL_CULL_FACE);
```
dethrace-labs, line 555: I would put this code in SetActiveMaterial.

And by framing it like this, we don’t have to have a separate piece of reset code.
if (double sided)
  glDisable(culling)
else
  glEnable(culling

zear: I thought about this initially, but the problem with doing so is that you can leave the culling disabled if the last model had `BR_MATF_TWO_SIDED`. So if anything else is rendered in the frame after the models, it might cause problems.

jeff-1amstudios:
I think that’s ok - either the next thing to be rendered is
- another model with its own materials which will setup the rendering
config with its own settings, - that model was the last one, and we move on
to rendering the framebuffer which in this case doesn’t care about the
culling, but generally should be setting the rendering config regardless of
what the current config is

src/harness/renderers/gl/gl_renderer.c
dethrace-labs: I think that’s ok - either the next thing to be rendered is
- another model with its own materials which will setup the rendering
config with its own settings, - that model was the last one, and we move on
to rendering the framebuffer which in this case doesn’t care about the
culling, but generally should be setting the rendering config regardless of
what the current config is

zear: Updated the patch to reflect the discussion.

src/DETHRACE/common/oil.c
```diff
@@ -41,6 +41,7 @@ void InitOilSpills() {
         the_material->flags |= BR_MATF_LIGHT;
         the_material->flags |= BR_MATF_PERSPECTIVE;
         the_material->flags |= BR_MATF_SMOOTH;
+        the_material->flags |= BR_MATF_TWO_SIDED;
```
dethrace-labs, line 44:
```suggestion
        // TODO: added by dethrace, investigate why oil spills in OG do not need this flag set to render correctly
        the_material->flags |= BR_MATF_TWO_SIDED;
```

src/harness/renderers/gl/gl_renderer.c
```diff
@@ -500,6 +500,11 @@ void setActiveMaterial(tStored_material* material) {
         glUniform1i(uniforms_3d.light_value, -1);
     }
 
+    if (material->flags & BR_MATF_TWO_SIDED)
+        glDisable(GL_CULL_FACE);
+    else
+        glEnable(GL_CULL_FACE);
+
```
dethrace-labs, line 507:
```suggestion
    if (material->flags & BR_MATF_TWO_SIDED || material->flags & BR_MATF_ALWAYS_VISIBLE) {
        glDisable(GL_CULL_FACE);
    } else {
        glEnable(GL_CULL_FACE);
    }
```

dethrace-labs: there is another flag `BR_MATF_ALWAYS_VISIBLE` that we should probably treat the same way

dethrace-labs (approved):
LGTM!

PR #242, opened 2022-11-11
Implement SmudgeCar

### LINKED ISSUE #241: SmudgeCar not implemented
https://github.com/dethrace-labs/dethrace/blob/4c9795a37b3a10833ecd447056b9d60290206b96/src/DETHRACE/common/spark.c#L1873

madebr:
Fixes #241

SmudgeCar modifies the "darkness" of vertices of the prepared model.
I'm not 100% confident the GL renderer will use these updates vertices since no `BrModelUpdate` is called.
Or the renderern does not respect the darkness.

### DIFF
--- a/src/DETHRACE/common/spark.c
+++ b/src/DETHRACE/common/spark.c
@@ -1887,7 +1887,91 @@ void SmudgeCar(tCar_spec* pCar, int fire_point) {
     int n;
     LOG_TRACE("(%p, %d)", pCar, fire_point);
 
-    STUB_ONCE();
+    if (gAusterity_mode) {
+        return;
+    }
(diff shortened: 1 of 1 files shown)


dethrace-labs (approved):

dethrace-labs:
the loops seem like they are duplicated - do you think the code was copy-pasted accidentally during development?

dethrace-labs:
> the loops seem like they are duplicated - do you think the code was copy-pasted accidentally during development?

nevermind - I see it is one loop for main model, one loop for the bonnet model 👍

dethrace-labs:
Added a separate issue for the gl renderer

PR #249, opened 2022-11-18
Fix for running with -robots + implement DOS gore password check

madebr:
- Running DethRace with `-robots` segfaulted because `strtok` modifies read-only data. I found this error weird, because non-zombie mode also modifies read-only data. Fix this by copying the read-only data to the stack before modification.
- The Carmageddon DOS version asks for a password to allow running DethRace with gore. Re-implement this behavior when starting DethRace with `--dos-mode`

### DIFF
--- a/src/DETHRACE/pc-dos/dossys.c
+++ b/src/DETHRACE/pc-dos/dossys.c
@@ -682,6 +682,34 @@ int OurGetChar() {
     NOT_IMPLEMENTED();
 }
 
+int CheckGorePasswordFile(char* pPassword) {
+    tPath_name path;
+    FILE* f;
+    char line[10];
+
+    PathCat(path, "DATA", "PASS.TXT");
+    f = fopen(path, "rb");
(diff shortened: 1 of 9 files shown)


src/DETHRACE/pc-dos/dossys.c
```diff
@@ -682,21 +682,61 @@ int OurGetChar() {
...
+        return 0;
+    }
+    fclose(f);
+#if defined(DETHRACE_FIX_BUGS)
+    // Allow a final newline in DATA/PASS.TXT
+    while (strlen(line) > 0 && (line[strlen(line) - 1] == '\n' || line[strlen(line) - 1] == '\r')) {
+        line[strlen(line) - 1] = '\0';
+    }
+#endif
+#ifdef _WIN32
```
dethrace-labs, line 706: Or add `#define strcasecmp stricmp` in windows.h ?

dethrace-labs (approved):

PR #252, opened 2022-11-22
Fix vehicle suspension so that it actually works

### LINKED ISSUE #221: Jelly Suspension makes wheels move, not the suspension
Maybe not a problem with the powerup itself, maybe a problem with how the wheels and suspension work that jelly suspension makes more obvious...?

The wheels are supposed to stay on the ground and the car body to float on the high bouncing suspension. Instead, the car body _and_ wheels appear to float around.

https://user-images.githubusercontent.com/78985374/197294676-6ee6752b-ae20-4bed-893a-147f4e5d3cff.mov

dethrace-labs:
Suspension was actually not working at all. An error was obvious in #221, but actually was a general suspension problem and not  a problem with jelly suspension powerup.

The main fix is inside the “straight” handling in PathGrooveBastard, with some unrelated fixes for “circular” that I noticed at the same time. Plus random float fixes as usual.

fixes #221  

https://user-images.githubusercontent.com/78985374/203441958-2a8acba6-c179-47b8-a21a-fd3edf99af0a.mov

### DIFF
--- a/src/DETHRACE/common/car.c
+++ b/src/DETHRACE/common/car.c
@@ -4280,8 +4280,8 @@ void MungeCarGraphics(tU32 pFrame_period) {
                         } else if (!gOn_me_wheels_start) {
                             gOn_me_wheels_start = the_time;
                         } else if (the_time - gOn_me_wheels_start > 500
-                                && (the_car->last_special_volume == NULL
-                                    || the_car->last_special_volume->gravity_multiplier == 1.0f)) {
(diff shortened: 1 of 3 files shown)


madebr (changes requested):
See https://github.com/dethrace-labs/dethrace/pull/253 for my suggestions.
This PR also fixes moving cranes in the magnachem maps, I was about to create an issue for this.

madebr (approved):

Issue #214, opened 2022-10-16
Opponent smoke is drawn incorrectly behind level geometry


rafalp:
Smoke is drawn behind ice on ice platform level:

<img width="1440" alt="Zrzut ekranu 2022-10-16 o 21 47 13" src="https://user-images.githubusercontent.com/750553/196055237-d755b825-c850-44ba-ad3e-e06279dd23ad.png">

rafalp:
Looks like this is not specific to transparency. Here's four race in Splat Pack, and smike is drawn behind all geometry:

<img width="1440" alt="Zrzut ekranu 2022-10-16 o 22 28 39" src="https://user-images.githubusercontent.com/750553/196056748-9cb75bd3-fd4c-4dbb-9251-8fc7f0a0d8ee.png">

<img width="1440" alt="Zrzut ekranu 2022-10-16 o 22 27 08" src="https://user-images.githubusercontent.com/750553/196056762-04e20c21-2eca-4fcd-a4da-3b639e330453.png">

zear:
This also impacts other effects, e.g. Electro-bastard Ray is only visible on top of car shadows in "Fridge Racer" (3rd race in original Carmageddon) and in the aforementioned "Drown in the Sumps" (4th race from Splat Pak).

zear:
"By the Sea Scythe" is another race where this happens for me. For the sake of completeness, my renderer details:
```
[INFO] GLRenderer_Init OpenGL vendor string: AMD
[INFO] GLRenderer_Init OpenGL renderer string: AMD Radeon Vega 11 Graphics (raven, LLVM 14.0.6, DRM 3.48, 6.0.10-arch2-1)
[INFO] GLRenderer_Init OpenGL version string: 4.6 (Compatibility Profile) Mesa 22.2.3
[INFO] GLRenderer_Init OpenGL shading language version string: 4.60
```

dethrace-labs:
This was fixed in #255

PR #255, opened 2022-12-01
Fix framebuffer effects

zear:
Screen effects such as smoke, sparks or electro-bastard ray need to be applied only after the 3D scene is rendered. Otherwise, they might end up rendered behind geometry on certain levels. Examples of such levels include "Fridge Racer" and "Drown in the Sumps", where said effects are rendered behind most geometry (except car shadows), and "Icing on the Quake", where they are obscured by translucent ice texture (see #214 for details).

Since my lack of access to symbolicated disassembly of the original game prevents me from checking whether the original game renders those effects before or after `BrZbSceneRenderEnd`, perhaps this is just a prelude to a bigger issue in the rendering pipeline.

With this change, smoke in "Drown in the Sumps" is now correctly rendered at the right Z position:
sshot

And it also doesn't get obscured by ice in "Icing on the Quake" anymore:
sshot2

### DIFF
--- a/src/DETHRACE/common/graphics.c
+++ b/src/DETHRACE/common/graphics.c
@@ -12,6 +12,7 @@
 #include "globvars.h"
 #include "globvrpb.h"
 #include "grafdata.h"
+#include "harness/hooks.h"
 #include "harness/os.h"
 #include "harness/trace.h"
 #include "init.h"
@@ -1615,6 +1616,7 @@ void RenderAFrame(int pDepth_mask_on) {
             ProcessTrack(gUniverse_actor, &gProgram_state.track_spec, gCamera, &gCamera_to_world, 1);
         }
         RenderSplashes();
+	Harness_Hook_FlushRenderer(); /* Dethrace. Flush buffers into memory. */
         RenderSmoke(gRender_screen, gDepth_buffer, gCamera, &gCamera_to_world, gFrame_period);
         RenderSparks(gRender_screen, gDepth_buffer, gCamera, &gCamera_to_world, gFrame_period);
         RenderProximityRays(gRender_screen, gDepth_buffer, gCamera, &gCamera_to_world, gFrame_period);


zear:
Updated the patch after a session with OG disassembly. The actual reason for the effects not rendering on certain tracks was because those tracks didn't use depth effects (darkness, fog), preventing the framebuffer/depthbuffer from being flushed. After moving `Harness_Hook_FlushRenderer` out of `DoDepthByShadeTable` and into the main rendering loop, the aforementioned levels seem to render correctly for me.

P.S. I'm not sure whether this should be wrapped into `DETHRACE_FIX_BUGS`, as it's a renderer-dependent quirk. I wonder if the OG Glide renderer had a similar workaround.

zear:
One extra issue I caught: Turns out that after applying the depth effects, you still need to re-flush the buffers after the blending pass of `ProcessTrack`. Without it, tracks that contain transparency (I think that's exclusive to Splat Pack?), such as "Icing on the Quake", still have depth issues for smoke & other effects. Plus, `RenderSplashes` adds actors that alter the depth buffer anyway.

After correcting for the above, the effects now display correctly on all the known tracks that exhibited the issue in question.

zear:
> I wonder if the OG Glide renderer had a similar workaround.

Allow me to answer my own question - Glide doesn't exhibit those issues, because all effects appear to be done in hardware.

In OG Glide, smoke is a model that can clip into other geometry (take a look at the wall on the left):
glide

And it also gets perfectly z-buffered:
glide2

Compare that to DethRace's GL renderer, where smoke "shimmering" around that palm tree trunk can be seen, because smoke effect is done in software:
dethrace

dethrace-labs:
Great! Actually now you mention it, I remember we had a related problem with tracks which had no depth :)

Yes, the glide executable moves some of those things (at least depth and smoke) into hardware. This is why the 3dfx version, while crisper and better framerate, doesnt _look_ quite the same as the original. Here, we are seeking, at least for now, to emulate the behavior of the original software renderer

dethrace-labs:
Agree - we don't need to wrap this in `DETHRACE_FIX_BUGS`. Anything that starts with `Harness_Hook_` is expected to be obviously not part of the original code

dethrace-labs (approved):

Issue #256, opened 2022-12-02
Gear images misaligned for cars with more than 8 gears


zear:
The game naively assumes that all cars have up to 8 gears (R, N and 1..6) and use HGEARS.PIX.
However, Suppressor has 11 gears and uses HGEARS4.PIX, a bitmap with a different height. This messes up gear mask calculation:
![gears](https://user-images.githubusercontent.com/510643/205399897-a9665418-be03-4ad1-9db3-565bc0c72853.png)
... and causes a buffer overflow on higher gears, where garbage is displayed instead of gear numbers.

This bug is also present in OG.

PR #257, opened 2022-12-02
Fix gear rendering for Suppressor

zear:
Fix misaligned gear mask and a buffer overflow for cars which use more than 8 gears, and thus pixelmaps other than HGEARS.PIX.

This issue exists because the gear rendering code assumes that the pixelmap with gear images contains 8 individual gear images. However, this is only true for HGEARS.PIX. In case of the Suppressor, which uses HGEARS4.PIX, there are 11 gears and so the mask is applied at the wrong offset.
Since both of the pixelmaps in question have gears that should clip at the height of 28, hardcode this number instead of trying to calculate it.

Another solution would be to simply force HGEARS4.PIX for all the cars, as those graphics are identical for both files (just with more gears in the latter one). I guess they really wanted to save every byte of memory in the OG, huh?

This fixes https://github.com/dethrace-labs/dethrace/issues/256:
gear2

### DIFF
--- a/src/DETHRACE/common/displays.c
+++ b/src/DETHRACE/common/displays.c
@@ -1160,15 +1160,26 @@ void DoInstruments(tU32 pThe_time) {
             } else {
                 gear = gCar_to_view->gear;
             }
+#if defined(DETHRACE_FIX_BUGS)
+/*
+ * The OG derives gear mask height of 28 by `gears_image->height / 8`, but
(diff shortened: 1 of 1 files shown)


dethrace-labs (approved):
Nice find and fix! I’d never noticed this!

dethrace-labs:
Maybe you could change the “>> 3” to a “/ 8” since you are touching those lines? It’s more likely that was what the original code looked like

zear:
Done.

Issue #182, opened 2022-09-27
Extra time blue counter displays invalid number sometimes


rafalp:
I've got electro bastard ray powerup and massacred a lot of peds, but blue counter of extra time didn't go over 27:

<img width="1440" alt="Zrzut ekranu 2022-09-27 o 22 21 01" src="https://user-images.githubusercontent.com/750553/192628428-f9781184-bce6-48cc-957b-8a8b8c0a5973.png">

PR #258, opened 2022-12-03
Fix time credit headup

zear:
Store the previous time credit in `gLast_time_credit_amount` instead of `gLast_time_credit_headup`. The latter is supposed to hold the headup slot number and gets overwritten with `NewTextHeadupSlot` index, effectively capping the time credits at the value of `+13`.

This fixes https://github.com/dethrace-labs/dethrace/issues/182.

### DIFF
--- a/src/DETHRACE/common/displays.c
+++ b/src/DETHRACE/common/displays.c
@@ -1439,9 +1439,9 @@ void AwardTime(tU32 pTime) {
     }
     gOld_times[0] = pTime;
     if (gLast_time_credit_headup >= 0 && (the_time - gLast_time_earn_time) < 2000) {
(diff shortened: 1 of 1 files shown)


dethrace-labs:
Nice find. Did you check the other references to gLast_time_credit_headup and gLast_time_credit_amount to make sure they aren’t transposed in other places?

dethrace-labs (approved):

zear:
Yes, `gLast_time_credit_headup` is only used in `AwardTime` and `ClearHeadups`, while `gLast_time_credit_amount` hasn't been used at all (that's how I realized the mixup).

madebr (approved):

Issue #259, opened 2022-12-03
INDUSTB type tracks don't render skidmarks 


zear:
This is an OG bug, where "Acid Reign" (INDUSTB1.TXT) and "Ramp Rampage" (INDUSTB2.TXT) races don't contain any skidmark material metadata, preventing skidmarks from being displayed. Only oil and blood marks show up in those races.

"Acid Reign" after performing some "donuts":
![skids_missing](https://user-images.githubusercontent.com/510643/205449485-0408a93e-b65a-4ce0-a1a2-d6b18e768361.png)

While one could argue this is a deliberate design choice, the race "MagnaChem Mayhem" (INDUSTB3.TXT) does contain skidmark information and allows them to be rendered, strongly hinting towards an overlook of the level designer in those two aforementioned races.

Whether there are other track sets with this issue is unknown.

PR #260, opened 2022-12-03
Fix rendering of skidmarks in INDUSTB tracks

zear:
The "Ramp Rampage" and "Acid Reign" tracks don't contain any skidmark material metadata, preventing skidmarks from being displayed in those races. As a workaround, use skidmark information extracted from "MagnaChem Mayhem", a variation of the same track design.

This fixes https://github.com/dethrace-labs/dethrace/issues/259:
skids_fixed

### DIFF
--- a/src/DETHRACE/common/world.c
+++ b/src/DETHRACE/common/world.c
@@ -2390,6 +2390,9 @@ void LoadTrack(char* pFile_name, tTrack_spec* pTrack_spec, tRace_info* pRace_inf
     int style;
     int cp_rect_w[2];
     int cp_rect_h[2];
+#if defined(DETHRACE_FIX_BUGS)
+    int skid_mark_cnt = 0;
+#endif
     tPath_name the_path;
     tPath_name general_file_path;
     char s[256];
@@ -2806,8 +2809,21 @@ void LoadTrack(char* pFile_name, tTrack_spec* pTrack_spec, tRace_info* pRace_inf
             strcat(str, ".MAT");
             material = LoadSingleMaterial(&gTrack_storage_space, str);
             pRace_info->material_modifiers[i].skid_mark_material = material;
+#if defined(DETHRACE_FIX_BUGS)
+            skid_mark_cnt++;
+#endif
         }
     }
+#if defined(DETHRACE_FIX_BUGS)
+    /* Display skidmarks even if the race has no specified skidmark material. */
+    if (!skid_mark_cnt && num_materials) {
+        LOG_WARN("Track %s has no valid skid mark material, setting the default one",
+                 pRace_info->track_file_name);
+        LoadSinglePixelmap(&gTrack_storage_space, "SKIDMARK.PIX");
+        material = LoadSingleMaterial(&gTrack_storage_space, "SKIDMARK.MAT");
+        pRace_info->material_modifiers[0].skid_mark_material = material;
+    }
+#endif
     for (i = num_materials; i < 10; ++i) {
         pRace_info->material_modifiers[i].car_wall_friction = 1.0;
         pRace_info->material_modifiers[i].tyre_road_friction = 1.0;


src/DETHRACE/common/world.c
```diff
@@ -2798,6 +2798,22 @@ void LoadTrack(char* pFile_name, tTrack_spec* pTrack_spec, tRace_info* pRace_inf
...
             pRace_info->material_modifiers[i].skid_mark_material = NULL;
+#if defined(DETHRACE_FIX_BUGS)
+            /*
+             * Fix OG bug, where "INDUSTB1.TXT" and "INDUSTB2.TXT" tracks don't
+             * define any skidmark materials, preventing skidmarks from being
+             * displayed in those races. As a workaround, use skidmark names
+             * for material #0 as found in "INDUSTB3.TXT" (another variation of
+             * this track design).
+             */
+            if (i == 0 &&
```
madebr, line 2809: Are there tracks where the first skid material is `NULL`, but have valid skid materials for later indices?
I'm asking because it would be nice to avoid hard coding track file names.

Can we set the default skid mark on some condition?

e.g. add this at the end of the for loop.
```c
if (pRace_info->material_modifiers[0].skid_mark_material == NULL) {
    LOG_WARN("Track %s has no valid first skid mark material, setting the default one", pRace_info->track_file_name);
    LoadSinglePixelmap(&gTrack_storage_space, "SKIDMARK.PIX");
    material = LoadSingleMaterial(&gTrack_storage_space, "SKIDMARK.MAT");
    pRace_info->material_modifiers[i].skid_mark_material = material;
}
```

zear: Checked all the tracks for CARMA and SPLATPAK. The only tracks that don't set skid material (`INDUSTB1.TXT`, `INDUSTB2.TXT` and `NETINDB.TXT`), lack it for all the material indices. So I went ahead and modified your suggestion, so that it only applies the workaround when all the skid materials are missing.

src/DETHRACE/common/world.c
```diff
@@ -2390,6 +2390,7 @@ void LoadTrack(char* pFile_name, tTrack_spec* pTrack_spec, tRace_info* pRace_inf
     int style;
     int cp_rect_w[2];
     int cp_rect_h[2];
+    int skid_mark_cnt = 0; /* Dethrace. */
```
madebr, line 2393:
```suggestion
#if defined(DETHRACE_FIX_BUGS)
    int skid_mark_cnt = 0; /* Dethrace. */
#endif
```

src/DETHRACE/common/world.c
```diff
@@ -2806,8 +2807,19 @@ void LoadTrack(char* pFile_name, tTrack_spec* pTrack_spec, tRace_info* pRace_inf
             strcat(str, ".MAT");
             material = LoadSingleMaterial(&gTrack_storage_space, str);
             pRace_info->material_modifiers[i].skid_mark_material = material;
+            skid_mark_cnt++; /* Dethrace. */
```
madebr, line 2810:
```suggestion
#if defined(DETHRACE_FIX_BUGS)
            skid_mark_cnt++; /* Dethrace. */
#endif
```

madebr (approved):

src/DETHRACE/common/world.c
```diff
@@ -2806,8 +2809,21 @@ void LoadTrack(char* pFile_name, tTrack_spec* pTrack_spec, tRace_info* pRace_inf
...
             material = LoadSingleMaterial(&gTrack_storage_space, str);
             pRace_info->material_modifiers[i].skid_mark_material = material;
+#if defined(DETHRACE_FIX_BUGS)
+            skid_mark_cnt++;
+#endif
         }
     }
+#if defined(DETHRACE_FIX_BUGS)
+    /* Display skidmarks even if the race has no specified skidmark material. */
+    if (!skid_mark_cnt && num_materials) {
```
dethrace-labs, line 2819: why can't we just check `if (pRace_info->material_modifiers[0].skid_mark_material == NULL) { ... `?

zear: We could, but my idea is to only apply this fix if none of the materials specify skidmarks, not just the first one.

dethrace-labs: Ok, got it

Issue #121, opened 2022-04-25
Mouse clicking the up and down arrows in menus should hide the hand


dethrace-labs:
In the original game, the hand appears to be hidden for a second while the up/down animation occurs. In dethrace, the hand is not hidden, so appears to be cut off by the menu rendering over the top.

![image](https://user-images.githubusercontent.com/78985374/165030215-02c5ad26-fba8-493e-b3d5-089a938c9816.png)

PR #261, opened 2022-12-03
Hide mouse cursor when scrolling driver info

zear:
Mouse cursor is supposed to be hidden for a moment when scrolling through the driver info screen.

This fixes https://github.com/dethrace-labs/dethrace/issues/121.


madebr (approved):
Thanks for fixing this!

Issue #267, opened 2022-12-10
Incorrect headup hole lookup


zear:
When searching for the slot to display the headups, Death Race uses the first reported empty position, disregarding that the given headup might be already in use at a further index.

In certain situations, this can lead to two headups of the same type overlayed on top of each other:
![credit_slot](https://user-images.githubusercontent.com/510643/206857192-61f604bd-14e6-40fb-9bde-0a35ec0a541e.png)

https://user-images.githubusercontent.com/510643/206857177-cfd0b230-5d8b-41ec-a252-6e2baed4d2a5.mp4

dethrace-labs:
Nice! I had noticed this myself lately (the problem, not the fix)

madebr:
Fixed in https://github.com/dethrace-labs/dethrace/pull/268.
Thanks!

PR #268, opened 2022-12-10
Fix headup hole lookup

zear:
When searching for a particular headup slot, don't return on the first empty hole, but rather try to find if the matching slot is in use.

This resolves a bug where two headups of the same type could end up displayed on top of each other: https://github.com/dethrace-labs/dethrace/issues/267

Fix confirmed with OG disassembly.


dethrace-labs (approved):
Nice! I had noticed this myself lately (the problem, not the fix)

PR #269, opened 2022-12-11
Fix gear rendering for standard resolution mode

zear:
A different gear mask is being used for standard and hi-res modes. Supply a hardcoded mask value for standard resolution and detect the graphics mode.

This supplements the change in 49d8e963f44b26d0f21debf1152d4d322bff3095.

As this change hardcodes data dependent values, it has been verified with the following datasets:
* Carmageddon: standard res + hires
* Carmageddon Demo: standard res
* Splat Pack: standard res + hires
* Splat Pack Demo: standard res + hires

### DIFF
--- a/src/DETHRACE/common/displays.c
+++ b/src/DETHRACE/common/displays.c
@@ -1057,6 +1057,7 @@ void DoInstruments(tU32 pThe_time) {
     int the_wobble_x;
     int the_wobble_y;
     int gear;
+    int gear_height; /* Added by dethrace. */
     double the_angle;
     double the_angle2;
     double sin_angle;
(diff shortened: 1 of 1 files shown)


src/DETHRACE/common/displays.c
```diff
@@ -1162,24 +1163,27 @@ void DoInstruments(tU32 pThe_time) {
...
  * this is only valid for HGEARS.PIX, which contains 8 gear images. Hardcoding
  * this number fixes gear rendering for cars using HGEARS4.PIX, which consists
  * of 11 gear images.
  */
-#define GEAR_HEIGHT 28
+#define GEAR_HEIGHT 16
+#define GEAR_HEIGHT_HIRES 28
 #else
 #define GEAR_HEIGHT ((int)gProgram_state.current_car.gears_image->height / 8)
+#define GEAR_HEIGHT_HIRES GEAR_HEIGHT
```
dethrace-labs, line 1175: did you mean to include this line?

zear: I did. For OG behavior, we need the mask in both resolutions to be derived from the `(int)gProgram_state.current_car.gears_image->height / 8)` calculation.

dethrace-labs: Ah, of course, I missed the #else

dethrace-labs (approved):

Issue #271, opened 2022-12-13
Cockpit view not displaying correctly


zear:
The cockpit view in Dethrace is currently not rendering correctly.
This can be split down into 3 separate issues:

**1. Camera**

The cockpit viewport (`gCamera_list[0]`) matrix never gets updated. Technically, this can be fixed by adding the following lines in graphics.c:
```
     old_camera_matrix = gCamera->t.t.mat;
     if (gMirror_on__graphics) {
         old_mirror_cam_matrix = gRearview_camera->t.t.mat;
+        BrMatrix34Copy(&gRearview_camera->t.t.mat, &gRearview_camera_to_world);
     }
+    BrMatrix34Copy(&gCamera->t.t.mat, &gCamera_to_world);
```
...however, an implementation derived from OG needs to be tracked down.
Perhaps `old_camera_matrix = gCamera->t.t.mat` is there for a related reason.

_Deathrace after applying the above workaround:_
![deathrace](https://user-images.githubusercontent.com/510643/207315769-bcb980d8-1f16-46d7-b8ef-006c0ab694db.png)

**2. Cockpit dashboard**

The cockpit dashboard bitmap only displays for the standard resolution and is missing in the hires mode. In OG, it is displayed for both resolutions.

**3. Background**

The background, aka. "sky" does not render while in cockpit view. This is because `ExternalSky` and `DepthEffectSky` explicitly return early if `cockpit_on == true`. However, in OG the sky is still rendered in standard resolution and only gets ignored for hires.

_OG sky in standard resolution:_
![og](https://user-images.githubusercontent.com/510643/207314153-8522e98b-e9d9-4e3a-b99a-53f34e697fe6.png)

_OG sky in hires mode:_
![og_hires](https://user-images.githubusercontent.com/510643/207314157-4b234a85-e598-444a-8e19-2b7a0499859a.png)

dethrace-labs:
Hi @zear! Thanks for #272. Are you actively working on any other part of the cockpit issues? If not, its something I'm interested in fixing, so I can take it.

zear:
@dethrace-labs I sunk a couple extra days into this, but nothing bigger came out of it. Feel free to take over.
Let me just dump some insights, so that you don't have to rediscover it yourself.

https://github.com/dethrace-labs/dethrace/blob/716b98c25a2717862c528b630c18655b247e1e68/src/DETHRACE/common/world.c#L2641-L2643
[...]

PR #272, opened 2022-12-13
Fix hires cockpit image display

zear:
Strip map height is a 16 bit value. Cast to `uint16_t *`, so that the upper 8 bits don't get lost.

This fixes cockpit image rendering for -hires mode, where said images have heights exceeding 255px.

This patch resolves part 2 of https://github.com/dethrace-labs/dethrace/issues/271.

### DIFF
--- a/src/DETHRACE/common/loading.c
+++ b/src/DETHRACE/common/loading.c
@@ -842,7 +842,7 @@ tS8* ConvertPixToStripMap(br_pixelmap* pThe_br_map) {
     temp_strip_image = BrMemAllocate(pThe_br_map->row_bytes * pThe_br_map->height, kMem_strip_image);
     current_size = 2;
 
-    *temp_strip_image = pThe_br_map->height;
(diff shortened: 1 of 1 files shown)


src/DETHRACE/common/loading.c
```diff
@@ -842,7 +842,7 @@ tS8* ConvertPixToStripMap(br_pixelmap* pThe_br_map) {
     temp_strip_image = BrMemAllocate(pThe_br_map->row_bytes * pThe_br_map->height, kMem_strip_image);
     current_size = 2;
 
-    *temp_strip_image = pThe_br_map->height;
+    *(uint16_t *)temp_strip_image = pThe_br_map->height;
```
madebr, line 845: We're trying to use BRender data types in the dethrace code base.
```suggestion
    *(br_uint16_t*)temp_strip_image = pThe_br_map->height;
```

madebr (approved):

Issue #282, opened 2023-01-21
Rendering artifacts at different viewport sizes


zear:
Deathrace exhibits multiple issues related to rendering the 3d part of the screen in non-full resolution (i.e. manipulating the viewport size with -/+ keys):

**1. Depth buffer**

When the `gRender_screen` is scaled such that it doesn't match dimensions of `gBack_screen`, the depth buffer's position and size differs from the color buffer, causing visible artifacts on the screen:

![sshot3](https://user-images.githubusercontent.com/510643/213887331-4714c267-4a43-4b9a-a41d-18bceeddfd8d.png)

**2. Blending**

The viewport height passed to the shader should be the one of the renderer, not the color buffer. Otherwise, geometry behind transparent materials is sampled at the wrong y offset:

![sshot1](https://user-images.githubusercontent.com/510643/213887507-91d53fc8-5f64-495e-9bcc-ab61e1b13f22.png)

**3. Blending in map mode**

When map mode is enabled, the scene is rendered thrice to a single frame. This breaks blending in GL renderer, which depends on a clear depth buffer. The result is fully opaque transparencies:

![sshot2](https://user-images.githubusercontent.com/510643/213887620-33e595c3-d34d-4647-afe1-b7de5d7c3e5d.png)

**4. External sky**

Depending on the screen size, something seems to obstruct different parts of the external sky:

https://user-images.githubusercontent.com/510643/213887149-ec42931d-8d5b-4891-949f-cbcfd60c602b.mp4

 Currently, I can't tell whether this is geometry, or the sky bitmap not rendering properly.

zear:
Parts 1-3 are now fixed in https://github.com/dethrace-labs/dethrace/pull/283.

zear:
Regarding part 4, it is the chunk of `ExternalSky` code responsible for filling in the background with uniform color around the sky pixelmap that is at fault. After a change to draw in white, everything becomes clear:

https://user-images.githubusercontent.com/510643/213915480-20f67a8c-1386-46a6-871a-388aaaa9bbc6.mp4

Now, the reason this happens is because the below code doesn't take `pRender_buffer->base_x`/`pRender_buffer->base_y` into account while calculating `DRPixelmapRectangleFill` offsets:

https://github.com/dethrace-labs/dethrace/blob/83ba373426385882fdf098ce363bb0cf6ea2e71c/src/DETHRACE/common/depth.c#L498-L506

This can be easily fixed by respecting said offsets:
```
[...]

PR #279, opened 2023-01-13
Implement cockpit view

dethrace-labs:
- Fixes cockpit camera direction and position
- Fixes skybox model and rendering used in cockpit mode
- Updates `BR_MODU_*` enum names from BRender 1.3 that seem to match better than the BRender 1.2 SDK names
- Added support for rendering with `BR_RSTYLE_EDGES` - useful for debugging
- Reverted to using the colour_buffer width/height/offset when reading the depth buffer which resolves 1 and 2 from #282

https://user-images.githubusercontent.com/78985374/218414238-e247dd6e-64f4-4412-889b-acc93c617ec6.mov

Issue #143, opened 2022-09-02
No need to draw everything 3 times


madebr:
https://github.com/dethrace-labs/dethrace/blob/b3bdbb26503ac435b8877ccff4a6418884640b0e/src/DETHRACE/common/graphics.c#L1607

This for-loop causes the scene to be drawn 1 to 3 times, depending whether the map mode is enabled or not.
The original executable has a loop here, but when I replace this line with a simple `if (1) {`, everything keeps working 100% fine.

dethrace-labs:
I noticed that too, so strange! No idea why the original executable does that. I've seen it in all the executables I've looked at

madebr:
It would be interesting to patch this in an executable and measure whether there is a difference in frame rate.

dethrace-labs:
Interestingly, in `VOODOO2C.exe`, its slightly different:

```
v27 = !gMap_mode || gSome_var_only_set_in_voodoo_rendering_mode ? 1 : 3
```

So if it is in voodoo mode, the triple rendering doesn't happen. Its weird though - why go to the trouble of updating the check instead of removing the loop entirely? 🤔  

/ @madebr

dethrace-labs:
Maybe they were working around a bug in the software renderer

madebr:
I looked a bit in `3DFX.EXE`, and found the `gSome_var_only_set_in_voodoo_rendering_mode` variale is used in only 1 other funtion: `ReinitialiseRenderStuff`.
```c
void ReinitialiseRenderStuff() {
    int x_diff;
    int y_diff;
    LOG_TRACE("()");

    if (gProgram_state.cockpit_on) {
        gProgram_state.current_render_left = gProgram_state.current_car.render_left[gProgram_state.cockpit_image_index];
        gProgram_state.current_render_top = gProgram_state.current_car.render_top[gProgram_state.cockpit_image_index];
        gProgram_state.current_render_right = gProgram_state.current_car.render_right[gProgram_state.cockpit_image_index];
        gProgram_state.current_render_bottom = gProgram_state.current_car.render_bottom[gProgram_state.cockpit_image_index];
    } else {
        if (gSome_var_only_set_in_voodoo_rendering_mode) {
            gProgram_state.current_render_top = 0;
            gProgram_state.current_render_left = 0;
            gProgram_state.current_render_right = gGraf_specs[gGraf_spec_index].total_width;
            gProgram_state.current_render_bottom = gGraf_specs[gGraf_spec_index].total_height;
[...]

PR #283, opened 2023-01-21
Viewport fixes

zear:
This is a collection of fixes for https://github.com/dethrace-labs/dethrace/issues/282.

### DIFF
--- a/src/harness/renderers/gl/gl_renderer.c
+++ b/src/harness/renderers/gl/gl_renderer.c
@@ -340,7 +340,7 @@ void GLRenderer_BeginScene(br_actor* camera, br_pixelmap* colour_buffer, br_pixe
     glViewport(colour_buffer->base_x, render_height - colour_buffer->height - colour_buffer->base_y, colour_buffer->width, colour_buffer->height);
     glUseProgram(shader_program_3d);
     glEnable(GL_DEPTH_TEST);
-    glUniform1i(uniforms_3d.viewport_height, colour_buffer->height);
+    glUniform1i(uniforms_3d.viewport_height, render_height);
 
     int enabled_clip_planes = 0;
     for (int i = 0; i < v1db.enabled_clip_planes.max; i++) {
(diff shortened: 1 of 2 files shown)


dethrace-labs:
#279 resolves 1 and 2 of the linked issues - different implementation for 1 but same for 2!

src/harness/renderers/gl/gl_renderer.c
```diff
@@ -389,6 +389,10 @@ void GLRenderer_EndScene() {
 
     generated_current_framebuffer_for_this_frame = 0;
     CHECK_GL_ERROR("GLRenderer_EndScene");
+
+    /* Fixes blending in map mode, which renders the scene thrice per frame. */
+    glBindFramebuffer(GL_FRAMEBUFFER, framebuffer_id);
```
dethrace-labs, line 394: this has the effect of breaking shadow rendering. Instead, I'm happy to wrap the for loop in `RenderAFrame` in a `#if DETHRACE_FIX_BUGS`, which should also resolve the issue?

I've looked and looked and can't see any reason for it. My guess is it was some leftover from a debugging session that made it into the retail build as an oversight

zear: Yes, the issue is gone if we only loop once. I also spent some time trying to understand the logic behind the loop and got nothing. So I'm fine with the workaround you're proposing.
Have we tried to hack CARMA95.EXE, so it only loops once, to see if anything breaks there?

dethrace-labs: Interestingly, in `VOODOO2C.exe`, its slightly different:

```
v27 = !gMap_mode || gSome_var_only_set_in_voodoo_rendering_mode ? 1 : 3
```

So if it is in voodoo mode, the triple rendering doesn't happen. Its weird though - why go to the trouble of updating the check instead of removing the loop entirely? 🤔  

/ @madebr

dethrace-labs: yes, I was curious again so just tried it (I updated the linked issue #143). Couldn't see anything broken :/

Issue #284, opened 2023-02-10
Sound sample NULL pointer dereferences


zear:
It is possible to trigger a NULL pointer dereference by muting sound while in main menu, which prevents sound samples from loading with the rest of game data. When the sound is then unmuted during gameplay, an attempt to play affected sound effects will crash the game. This is most prominent for pratcam sounds.

This issue also occurs in OG and can be triggered in the Windows builds. It most likely has been carried over from the DOS codebase, but hasn't been discovered during original playtesting, since DOS doesn't have the same concept of memory protection as multi-tasking OSes.

Steps to reproduce:
1. Launch the game.
2. While in main menu, press `s` in order to mute sound.
3. Load a track all the way through into gameplay.
4. While in-game, press `s` again to unmute sounds.
5. Trigger a pratcam sound effect (e.g. by running over peds).

zear:
Fixed in https://github.com/dethrace-labs/dethrace/pull/285.

PR #285, opened 2023-02-10
Fix NULL pointer dereference in S3StartSound (#284)

zear:
If the sound has been muted (!gS3_enabled) during game data loading, it is then possible to unmute it during gameplay and request a sound effect which doesn't have a sample loaded (most prominent for pratcam sfx). While the code accommodates for such a case, it does it too late, triggering a NULL pointer dereference on platforms which feature memory protection. This bug is most likely an overlook from the DOS era.

Fix this by moving the sample NULL pointer check before its first use.

Signed-off-by: Artur Rojek <contact@artur-rojek.eu>

### DIFF
--- a/src/S3/CMakeLists.txt
+++ b/src/S3/CMakeLists.txt
@@ -18,6 +18,9 @@ else()
         /wd4996
     )
 endif()
+if(DETHRACE_FIX_BUGS)
+    target_compile_definitions(s3 PRIVATE DETHRACE_FIX_BUGS)
+endif()
 
 if(IS_BIGENDIAN)
     target_compile_definitions(s3 PRIVATE BR_ENDIAN_BIG=1)
--- a/src/S3/audio.c
+++ b/src/S3/audio.c
@@ -962,6 +962,13 @@ void S3CalculateRandomizedFields(tS3_channel* chan, tS3_descriptor* desc) {
     chan->left_volume = vol;
     chan->right_volume = vol;
     if (desc->type == eS3_ST_sample) {
+#if defined(DETHRACE_FIX_BUGS)
+    /* Avoid a possible NULL pointer dereference. */
+    if (desc->sound_data == NULL) {
+        chan->rate = desc->min_pitch;
+        return;
+    }
+#endif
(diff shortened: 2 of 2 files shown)


src/S3/audio.c
```diff
@@ -793,6 +793,18 @@ tS3_sound_tag S3StartSound(tS3_outlet* pOutlet, tS3_sound_id pSound) {
         gS3_last_error = eS3_error_bad_id;
         return 0;
     }
+#if defined(S3_FIX_BUGS)
```
dethrace-labs, line 796: I think we could use `DETHRACE_FIX_BUGS` here - its probably unlikely someone would want to fix bugs in one area but not the other

src/S3/audio.c
```diff
@@ -793,6 +793,18 @@ tS3_sound_tag S3StartSound(tS3_outlet* pOutlet, tS3_sound_id pSound) {
...
+/*
+ * Load the sample early to prevent a possible NULL pointer dereference in
```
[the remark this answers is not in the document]
zear: Done. And I can't even hear a difference between `chan->rate = desc->min_pitch;` and `chan->rate = S3IRandomBetweenLog(...);` in practice :)

src/S3/CMakeLists.txt
```diff
@@ -18,6 +18,9 @@ else()
         /wd4996
     )
 endif()
+if(DETHRACE_FIX_BUGS)
```
madebr, line 21: Not in this bug report, but I think we should combine `BRENDER_FIX_BUGS` and `DETHRACE_FIX_BUGS` in one cmake option + compile definition.

madebr (approved):
I can reproduce the crash in OG and current dethrace.

Just a little question. Would always loading the samples, but not playing them, when sound is disabled be much harder to implement?

dethrace-labs (approved):
LGTM

dethrace-labs:
@zear I've invited you as a collaborator to the project. If you accept, you will be able to merge your own approved PRs when you are ready. And thanks!

madebr:
Welcome to the 🚗  club @zear !

PR #294, opened 2023-02-21
Implements quitgame function

dethrace-labs:
- `QuitGame` was only partially implemented
- Added extra S3 functions required for shutting down
- Moved S3 device opening functions to something more closely matching OG code

### DIFF
--- a/docs/PORTING.md
+++ b/docs/PORTING.md
@@ -38,7 +38,7 @@ To add a new `IOPlatform`:
 - `Window_Create`
 - `Window_PollEvents`
 - `Window_Swap`
-- `Window_Close`
+- `IOPlatform_Shutdown`
 - `Input_GetKeyMap`
 - `Input_IsKeyDown`
 
--- a/docs/PORTING.md
+++ b/docs/PORTING.md
@@ -35,14 +35,17 @@ The default IO platform is `SDL_OpenGL`, which uses SDL for windowing and input,
 To add a new `IOPlatform`:
 
 1. Create `io_platforms/my_platform.c` file and implement the required functions defined in [io_platforms/io_platform.h](https://github.com/dethrace-labs/dethrace/blob/main/src/harness/io_platforms/io_platform.h):
-- `Window_Create`
-- `Window_PollEvents`
-- `Window_Swap`
+- `IOPlatform_Init`
+- `IOPlatform_CreateWindow`
+- `IOPlatform_PollEvents`
+- `IOPlatform_SwapWindow`
+- `IOPlatform_GetKeyMap`
+- `IOPlatform_IsKeyDown`
+- `IOPlatform_GetMousePosition`
+- `IOPlatform_GetMouseButtons`
 - `IOPlatform_Shutdown`
-- `Input_GetKeyMap`
-- `Input_IsKeyDown`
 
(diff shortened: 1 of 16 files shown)


src/DETHRACE/common/main.c
```diff
@@ -27,7 +29,21 @@ void QuitGame() {
...
+    ShutdownNetIfRequired();
+    if (gSound_available) {
+        DRS3ShutDown();
+    }
+    if (gBr_initialized) {
+        ClearEntireScreen();
+    }
+    PDRevertPalette();
+    StopMusic();
+    PDShutdownSystem();
```
madebr, line 46: I can also see `CloseDiagnostics()` and `exit(0)`, even though `PDShutdownSystem()` on Windows will already exit the system.
But I'm not sure
```suggestion
    PDShutdownSystem();
    CloseDiagnostics();
    exit(0);
```

src/DETHRACE/pc-dos/dossys.c
```diff
@@ -308,7 +308,8 @@ void PDInitialiseSystem() {
 // IDA: void __cdecl PDShutdownSystem()
 void PDShutdownSystem() {
     LOG_TRACE("()");
-    NOT_IMPLEMENTED();
+
+    exit(0);
```
madebr, line 312: Not needed now, but perhaps we need to add a harness hook here?
We can then destroy the OpenGL and SDL contexts.

dethrace-labs: good point, yes, added it 👍

src/DETHRACE/common/main.c
dethrace-labs: good spotting! For some reason, IDA was confused by this function and the disassembler was missing the last bit of it. Good reminder to not be lazy and double check the assembly too :)

madebr: In that regard, ghidra is moody too.
It does not always show what happens after `FatalError`, but the decompiler showed everything here.
I suppose you need to always be careful when marking a function as "does not return".

src/harness/io_platforms/sdl_gl.c
```diff
@@ -203,6 +203,12 @@ tRenderer* Window_Create(char* title, int width, int height, int pRender_width,
     return &gl_renderer;
 }
 
+void Window_Close() {
+    SDL_GL_DeleteContext(context);
+    SDL_DestroyWindow(window);
+    window = NULL;
```
madebr, line 209: You can also do `SDL_Quit()`.

src/harness/harness.c
```diff
@@ -317,6 +317,10 @@ void Harness_Hook_GraphicsInit(int render_width, int render_height) {
     renderer = Window_Create("Dethrace", window_width, window_height, render_width, render_height);
 }
 
+void Harness_Hook_PDShutdownSystem() {
+    Window_Close();
```
madebr, line 321: The platform also must shut down, in this case SDL.

docs/PORTING.md
```diff
@@ -38,6 +38,7 @@ To add a new `IOPlatform`:
 - `Window_Create`
 - `Window_PollEvents`
 - `Window_Swap`
+- `IOPlatform_Shutdown`
 - `Input_GetKeyMap`
 - `Input_IsKeyDown`
```
madebr, line 43: Should all these functions have a `IOPlatform_` prefix? Right now it's a mishmash. (something to think about)

madebr (approved):

Issue #132, opened 2022-06-03
segfault when crashing with smoke against palm tree in Mayan Mayhem


madebr:
When crashing against a palm tree with a smoking (=broken) car in Mayan Mayhem (first map of Carmageddon Splat Pack), the game segfaults.

The following stack trace is print:
```
0: SmokeLine at /home/maarten/programming/dethrace/src/DETHRACE/common/spark.c:846
1: SmokeCircle at /home/maarten/programming/dethrace/src/DETHRACE/common/spark.c:949
2: SmokeCircle3D at /home/maarten/programming/dethrace/src/DETHRACE/common/spark.c:1084
3: DrawTheGlow at /home/maarten/programming/dethrace/src/DETHRACE/common/spark.c:1837 (discriminator 2)
4: RenderSmoke at /home/maarten/programming/dethrace/src/DETHRACE/common/spark.c:1176
5: RenderAFrame at /home/maarten/programming/dethrace/src/DETHRACE/common/graphics.c:1620 (discriminator 2)
6: MainGameLoop at /home/maarten/programming/dethrace/src/DETHRACE/common/mainloop.c:596
7: DoRace at /home/maarten/programming/dethrace/src/DETHRACE/common/mainloop.c:722
8: DoGame at /home/maarten/programming/dethrace/src/DETHRACE/common/structur.c:538
9: DoProgram at /home/maarten/programming/dethrace/src/DETHRACE/common/structur.c:646
10: GameMain at /home/maarten/programming/dethrace/src/DETHRACE/common/main.c:106
11: original_main at /home/maarten/programming/dethrace/src/DETHRACE/pc-dos/dossys.c:666
12: main at /home/maarten/programming/dethrace/src/DETHRACE/main.c:29
13: ?? ??:0
```

**How to reproduce**
- Start DethRace with `DETHRACE_ROOT_DIR` pointing to a splat pack directory
- Start a new game, select *Max Damage*, any difficulty, and start the *Mayan Mayhem* race
- When in-race, Enter `IBETYOUCANTPRINTCUNT` for entering the edit mode
- Press `F4` until you enter "damage edit mode"
- Make speed (but not too much to not trigger #131), drive a bit around in the grass and crash against a palm tree. Before you crash, make sure you have a damaged car that generates smoke. Use the the keys `0`-`6` (combined with `SHIFT`) for that.

dethrace-labs:
[...]

PR #295, opened 2023-02-26
Fix buffer under/overflows in `SmokeLine` (#132)

zear:
When `SmokeLine` is called through `DrawTheGlow`, the calculated shade table offset might result in a negative value. Subject to usual arithmetic conversions, it is then treated as a large unsigned array index. On 32-bit systems (e.g. OG builds for DOS, Windows 95), the pointer arithmetic overflows and produces a negative index access, simply grabbing game data a few bytes before the start of the table. On 64-bit platforms this instead results in page fault.

1. Keep table offset in a signed integer, making the negative values explicit. This provides OG behavior for 64-bit builds.
2. Optionally, cap all negative values at 0, preventing underflows.

### DIFF
--- a/src/DETHRACE/common/spark.c
+++ b/src/DETHRACE/common/spark.c
@@ -858,6 +858,7 @@ void DrMatrix34Rotate(br_matrix34* mat, br_angle r, br_vector3* a) {
 // IDA: void __usercall SmokeLine(int l@<EAX>, int x@<EDX>, br_scalar zbuff, int r_squared, tU8 *scr_ptr, tU16 *depth_ptr, tU8 *shade_ptr, br_scalar r_multiplier, br_scalar z_multiplier, br_scalar shade_offset)
 void SmokeLine(int l, int x, br_scalar zbuff, int r_squared, tU8* scr_ptr, tU16* depth_ptr, tU8* shade_ptr, br_scalar r_multiplier, br_scalar z_multiplier, br_scalar shade_offset) {
     int i;
+    int offset; /* Added by dethrace. */
     int r_multiplier_int;
     int shade_offset_int;
     tU16 z;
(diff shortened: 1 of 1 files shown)


src/DETHRACE/common/spark.c
```diff
@@ -858,6 +858,7 @@ void DrMatrix34Rotate(br_matrix34* mat, br_angle r, br_vector3* a) {
 // IDA: void __usercall SmokeLine(int l@<EAX>, int x@<EDX>, br_scalar zbuff, int r_squared, tU8 *scr_ptr, tU16 *depth_ptr, tU8 *shade_ptr, br_scalar r_multiplier, br_scalar z_multiplier, br_scalar shade_offset)
 void SmokeLine(int l, int x, br_scalar zbuff, int r_squared, tU8* scr_ptr, tU16* depth_ptr, tU8* shade_ptr, br_scalar r_multiplier, br_scalar z_multiplier, br_scalar shade_offset) {
     int i;
+    int offset;
```
dethrace-labs, line 861:
```suggestion
    int offset;  // added by dethrace
```

dethrace-labs (approved):

src/DETHRACE/common/spark.c
```diff
@@ -873,7 +874,13 @@ void SmokeLine(int l, int x, br_scalar zbuff, int r_squared, tU8* scr_ptr, tU16*
 
     for (i = 0; i < l; i++) {
         if (*depth_ptr > z) {
-            *scr_ptr = shade_ptr[*scr_ptr + (((shade_offset_int - r_squared * r_multiplier_int) >> 8) & 0xffffff00)];
+            offset = ((shade_offset_int - r_squared * r_multiplier_int) >> 8) &
+		     0xffffff00;
+#if defined(DETHRACE_FIX_BUGS)
+            /* Prevent buffer underflows by capping negative offsets. */
+            offset = offset < 0 ? 0 : offset;
```
madebr, line 881: At this point, we're bike shedding :)
```suggestion
            offset = MAX(0, offset);
```

madebr (approved):

src/DETHRACE/common/spark.c
zear: I don't mind. A good patch is a clean patch.

Issue #297, opened 2023-02-27
Driving in grass emits ugly brown smoke


madebr:
Creating smoke when driving in the grass is ugly. It's plain brown instead of a transparent color.

This happens on Dethrace, using a Linux x86_64 system.

Actual behavior (Dethrace):
<video src='https://user-images.githubusercontent.com/4138939/221445536-299446b0-967d-4945-bd2a-fcf02ad21108.mp4' width=180/>

Expected behavior (DOS Carmageddon):
<video src='https://user-images.githubusercontent.com/4138939/221445593-963c9945-8145-4367-a239-b0b23032d1bf.mp4' width=180/>

zear:
Can't replicate that here (also Linux x86_64), the smoke color looks exactly as in OG.
That said, 12e941081a8216cd0bcfc2fad184a876967e2f98 touched code related to the smoke shade palette offsets - can you remove that patch on your end and see if that resolves the issue?

madebr:
This ugly smoke has always been present on my system since it was added to the code.
was just never able to find the cause.

The brown smoke is also present in the Windows dethrace executable:
![windows-dethrace](https://user-images.githubusercontent.com/4138939/221582827-fbbbdded-8175-43dd-ab6f-94778d3428bb.png)

It is not present in the OG Windows splatpack executable:
![windows-carma95](https://user-images.githubusercontent.com/4138939/221582899-94ee4401-1ab6-4e30-96ee-82de657880d0.png)

madebr:
Exact same executable creates "good" smoke when using german game data:
![DUMP0000](https://user-images.githubusercontent.com/4138939/222937061-ea382845-b06c-4ae3-b66b-4f592192725a.PNG)

The smoke of damaged/wasted cars is also working with the german game data:

![DUMP0025b](https://user-images.githubusercontent.com/4138939/222937334-e0875731-5185-4361-a810-61cd5809d2b3.PNG)

![DUMP0000b](https://user-images.githubusercontent.com/4138939/222937364-5439cb7f-8d85-4491-8762-bf148377a077.PNG)

madebr:
I found the issue by comparing the german game data with the english one.
There were corrupted `st[A-P]` files in my `DATA/SHADETAB` folder, probably from when I was implementing `GenerateShadeTable`.
Removing these files and let Dethrace regenerate them fixed everything.

PR #298, opened 2023-03-06
utility: fix filename of generated shade table

madebr:
The generated shade tables in `DATA/SHADETAB` were missing a `.TAB` suffix.
Found while looking into #297.


zear:
LGTM

zear (approved):

PR #300, opened 2023-03-31
Smudged vertices

### LINKED ISSUE #243: GL Renderer does not respect vertex colors
When creating the VBOs, we should be including the i8 `index` portion of the `vertex_colour` field so it can be used in the shaders

Ref: #242

dethrace-labs:
- Implements per-vertex coloring (`BR_MATF_PRELIT` materials). 
- Removes a hack to set a hardcoded default actor material by correctly using the parent material.

The burnt paint effect after being on fire relies on `BR_MATF_PRELIT`.

`SmudgeCar` picks vertices near the fire vertex and darkens them by setting the alpha component of the vertex color to a number 0-255 (higher values are darker).

These numbers are used in the renderer to pick a lit value per pixel from the associated shadetable.

Screenshot 2023-03-31 at 8 51 05 pm

Fixes #243

### DIFF
--- a/src/DETHRACE/common/spark.c
+++ b/src/DETHRACE/common/spark.c
@@ -875,8 +875,7 @@ void SmokeLine(int l, int x, br_scalar zbuff, int r_squared, tU8* scr_ptr, tU16*
 
     for (i = 0; i < l; i++) {
         if (*depth_ptr > z) {
-            offset = ((shade_offset_int - r_squared * r_multiplier_int) >> 8) &
-                     0xffffff00;
+            offset = ((shade_offset_int - r_squared * r_multiplier_int) >> 8) & 0xffffff00;
 #if defined(DETHRACE_FIX_BUGS)
             /* Prevent buffer underflows by capping negative offsets. */
             offset = MAX(0, offset);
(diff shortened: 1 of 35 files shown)


src/DETHRACE/constants.h
```diff
@@ -491,7 +491,7 @@ enum {
 
 #define TIME_CONV_THING 0.0005f
 
-#define OPPONENT_COUNT 0
+#define OPPONENT_COUNT 5
```
madebr: Haha, perhaps a `--opponent-count` argument option could be introduced?

dethrace-labs: haha yeah, I've thought the same a few times and havent got around to it :)

src/DETHRACE/common/spark.c
```diff
@@ -2051,84 +2051,92 @@ void SmudgeCar(tCar_spec* pCar, int fire_point) {
...
-    j = 0;
+    real_vertex_number = 0;
     if ((model->flags & BR_MODF_KEEP_ORIGINAL) != 0 || (model->flags & BR_MODF_UPDATEABLE) != 0) {
-        BrVector3Copy(&bonny_pos, &V11MODEL(model)->groups->vertices[fire_point].p);
+        point = V11MODEL(model)->groups[group].vertices[fire_point].p;
         StartPipingSession(ePipe_chunk_smudge);
-        for (group = 0; group < V11MODEL(model)->ngroups; group++) {
-            for (v = 0; v < V11MODEL(model)->groups[group].nvertices; v++) {
-                BrVector3Sub(&tv, &V11MODEL(model)->groups[group].vertices[v].p, &bonny_pos);
+        for (group = 0; V11MODEL(model)->ngroups > group; group++) {
```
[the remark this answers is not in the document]
dethrace-labs: thanks, I missed that one

PR #301, opened 2023-04-14
Implements win95sys.c

dethrace-labs:
The OG code already has a split between `common` and `pd` (platform-dependent) code. The original work was done on the DOS release which is where the platform-dependent code currently was pulled from.

However, we are now using the windows release, where platform dependencies are better defined - namely small portions of the windows API and small portions of DirectInput and DirectDraw.

This PR switches us to us the OG win95 platform code, with a small [shim](src/harness/win95/polyfill.c) implementing the Windows API in a cross-platform way. This allows the code in `win95sys.c` to be as close to OG as possible


src/DETHRACE/common/graphics.c
```diff
@@ -2574,10 +2574,10 @@ int DoMouseCursor() {
         period = 1000;
     }
     while (period <= 20) {
-        // Sleep 1 ms to avoid 100% CPU usage
-        OS_Sleep(1);
         this_call_time = PDGetTotalTime();
         period = this_call_time - last_call_time;
+        // added by dethrace to avoid 100% CPU usage
+        gHarness_platform.Sleep(1);
     }
```
madebr, line 2581: I think it's better to sleep first. Because now the time is sampled before the sleep which might cause an extra sleep.

PR #302, opened 2023-04-16
Implements drugs powerup

dethrace-labs:
The "Drugs!" powerup is implemented in two different parts

- `TheEffectsOfDrugs` function which runs every frame that shuffles palette colors around  (already done)
- `DoDrugWobbleCamera` function which manipulates the camera matrix to make the 3d scene 'wobble'

Interestingly, `TheEffectsOfDrugs` only changes the first 224 palette entries, both in DOS and Windows releases. This is why in snowy maps, the ground (being mostly white, palette index 255) is unaffected.

Screenshot 2023-04-17 at 10 06 19 am


madebr:
I submitted this pr to https://www.dea.gov/submit-tip :rofl:

PR #303, opened 2023-04-18
SmackerLib API

dethrace-labs:
Implements a small shim exposing the [RAD Smacker API](https://wiki.multimedia.cx/index.php/RAD_Game_Tools_Smacker_API) to `dethrace` so we can keep the code in `cutscene.c` as original as possible. 

libsmacker is now only a hidden dependency of smackw32


src/DETHRACE/common/cutscene.c
```diff
@@ -59,21 +59,13 @@ void PlaySmackerFile(char* pSmack_name) {
     tPath_name the_path;
     br_colour* br_colours_ptr;
     tU8* smack_colours_ptr;
-    // Smack* smk;
```
madebr, line 62: FYI, I was able to reverse the logic for carmageddon 2: https://github.com/madebr/rec2/blob/master/src/carma2/common/cutscene.c

dethrace-labs: awesome. So much similarity. I'm sad they got rid of the `fuck_off` variable :P

dethrace-labs: nevermind, I forgot you have to make up your own names. Well, probably what you have as `keyPressed` is really still `fuck_off` in the OG code! :)

PR #311, opened 2023-04-26
Remove references to `u_material_index_range`

### LINKED ISSUE #304: glGetUniformLocation(program, "u_material_index_range") fails
`dethrace` of current master fails with the following message:
```
[PANIC] GetValidatedUniformLocation glGetUniformLocation(4, u_material_index_range) failed. Check the shader uniform names.
```
I think what's happening is that the opengl shader compiler optimizes away unused uniforms.
Looking at the fragment shader, `u_material_index_range` is practically unused.
https://github.com/dethrace-labs/dethrace/blob/6c0b26cc8d72c21d4152adee3328aa88a3c95363/src/harness/resources/3d_frag.glsl#L46

Reading [this StackOverflow answer](https://stackoverflow.com/questions/21016310/disable-glsl-compiler-optimization), it looks like this optimization is spec compliant.

Because it's unused anyway, I locally commented out
https://github.com/dethrace-labs/dethrace/blob/6c0b26cc8d72c21d4152adee3328aa88a3c95363/src/harness/renderers/gl/gl_renderer.c#L162
and
https://github.com/dethrace-labs/dethrace/blob/6c0b26cc8d72c21d4152adee3328aa88a3c95363/src/harness/renderers/gl/gl_renderer.c#L549

dethrace-labs:
Removes unused reference to `u_material_index_range` uniform since it can be optimized away and cause `glGetUniformLocation` to fail.

Fixes #304

### DIFF
--- a/src/harness/renderers/gl/gl_renderer.c
+++ b/src/harness/renderers/gl/gl_renderer.c
@@ -161,6 +161,8 @@ void LoadShaders() {
     uniforms_3d.material_blend_enabled = GetValidatedUniformLocation(shader_program_3d, "u_material_blend_enabled");
     uniforms_3d.material_blend_table = GetValidatedUniformLocation(shader_program_3d, "u_material_blend_table");
     uniforms_3d.material_index_base = GetValidatedUniformLocation(shader_program_3d, "u_material_index_base");
+    // TODO: re-add when we support untextured lighting
+    // uniforms_3d.material_index_range = GetValidatedUniformLocation(shader_program_3d, "u_material_index_range");
 
     // bind the uniform samplers to texture units
     glUniform1i(uniforms_3d.material_texture_pixelmap, 0);
--- a/src/harness/renderers/gl/gl_renderer.c
+++ b/src/harness/renderers/gl/gl_renderer.c
@@ -65,7 +65,7 @@ struct {
     GLuint material_blend_enabled;
     GLuint material_blend_table;
     GLuint material_index_base;
-    GLuint material_index_range;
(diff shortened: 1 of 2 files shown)


src/harness/resources/3d_frag.glsl
```diff
@@ -66,7 +66,6 @@ void main() {
 
         if ((u_material_flags & BR_MATF_LIGHT) != 0u) {
             // TODO: lighting calculations based on https://rr2000.cwaboard.co.uk/R4/BRENDER/TEBK_43.HTM#0
-            uint range = u_material_index_range;
```
madebr, line 69: You can keep it here. The glsl compiler is not spewing warnings.

src/harness/renderers/gl/gl_renderer.c
```diff
@@ -161,7 +161,6 @@ void LoadShaders() {
     uniforms_3d.material_blend_enabled = GetValidatedUniformLocation(shader_program_3d, "u_material_blend_enabled");
     uniforms_3d.material_blend_table = GetValidatedUniformLocation(shader_program_3d, "u_material_blend_table");
     uniforms_3d.material_index_base = GetValidatedUniformLocation(shader_program_3d, "u_material_index_base");
-    uniforms_3d.material_index_range = GetValidatedUniformLocation(shader_program_3d, "u_material_index_range");
```
madebr, line 164: Add a fixme instead of removing.

src/harness/resources/3d_frag.glsl
dethrace-labs: all good, this line was only here because it stopped `u_material_index_range` being optimized away on OSX :)

PR #313, opened 2023-04-29
Provide an argument to GLRenderer_FlushBuffer

### LINKED ISSUE #310: No smoke is drawn in right half of screen
This happens in cockpit mode (320x200):
DUMP0025
Also in hi-res mode (640x400):
DUMP0025b
This also happens with smoke emitted by opponents:
DUMP0025c

Tested on current main: 35a4ce83dfb03e10f504416117410c4e381e90f8

dethrace-labs:
Part of the `tHarness_platform` set of function pointers is `void (*Renderer_FlushBuffers)();`. It does not take any arguments. We were assigning `GLRenderer_FlushBuffers` to that, which took a `tRenderer_flush_type` argument.

Because we were not providing the argument, its value was undefined and sometimes would appear to work and sometimes not.

fixes #310

### DIFF
--- a/src/harness/platforms/sdl_opengl.c
+++ b/src/harness/platforms/sdl_opengl.c
@@ -219,6 +219,10 @@ void set_palette(PALETTEENTRY_* pal) {
     GLRenderer_SetPalette((uint8_t*)pal);
 }
 
+void flush_buffers() {
+    GLRenderer_FlushBuffers(eFlush_all);
+}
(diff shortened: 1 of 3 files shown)


dethrace-labs:
@madebr could you see if this resolves #310 for you?

madebr:
Indeed! This fixes both issues mentioned in #310!

PR #318, opened 2023-05-01
Adds (void) to all functions taking no args

dethrace-labs:
This was necessary to make the latest `main` compile in my OSX environment (clang-14.0.3).

Otherwise, I was seeing ``` error: a function declaration without a prototype is deprecated in all versions of C [-Werror,-Wstrict-prototypes]``` for all void functions

### DIFF
--- a/src/BRSRC13/CORE/DOSIO/eventq.c
+++ b/src/BRSRC13/CORE/DOSIO/eventq.c
@@ -5,7 +5,7 @@ int _DOSEventQEnabled;
 host_real_memory _DOSEventMemory;
 
 // IDA: br_error __cdecl DOSEventBegin()
-br_error DOSEventBegin() {
+br_error DOSEventBegin(void) {
     host_info hi;
     int s;
     LOG_TRACE("()");
@@ -13,7 +13,7 @@ br_error DOSEventBegin() {
 }
 
 // IDA: void __cdecl DOSEventEnd()
-void DOSEventEnd() {
+void DOSEventEnd(void) {
     LOG_TRACE("()");
     NOT_IMPLEMENTED();
 }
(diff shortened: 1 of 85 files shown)


madebr:
I think this is the wrong approach to fix the error.
We need to be careful to declare public symbols in headers, declare private symbols in the source file, and add `static` to them. ODR for C.
I did [something similar for private symbols in the `-Wstrict-prototypes` pr](https://github.com/dethrace-labs/dethrace/blob/33f682ce8210ff0d3fc132dd01cdfa5977977e40/src/DETHRACE/pc-win95/win95sys.c#L82-L105).
We should also get rid of all `extern` function declarations in c sources, [such as here](https://github.com/dethrace-labs/dethrace/blob/33f682ce8210ff0d3fc132dd01cdfa5977977e40/src/S3/audio.c#L20). These declarations can get out of sync.

It looks like the warnings you're seeing are warnings emitted by `-Wmissing-declarations` on gcc

>       -Wmissing-declarations
>           Warn if a global function is defined without a previous declaration.  Do so even if the definition
>           itself provides a prototype.  Use this option to detect global functions that are not declared in header
>           files.  In C, no warnings are issued for functions with previous non-prototype declarations; use
>           -Wmissing-prototypes to detect missing prototypes.  In C++, no warnings are issued for function
>           templates, or for inline functions, or for functions in anonymous namespaces.

example:
the following example will emit the warning you're seeing.
The fix is to add `#include <a.h>` to `a.c`.
That way we are sure `a.h` and `a.c` remain in sync.

**a.h**
```c
int my_function(void);
```

**a.c**
```c
// no #include <a.h>
int my_function() {
  return arg + 2;
}
```

dethrace-labs:
Right, yep, agree with your points 👍 

But take a look at `BRSRC/DOSIO/eventq.c|h` (this was the first file that started complaining).
 - the .c does include the .h

the .h file contains `br_error DOSEventBegin(void);`, while the .c file has br_error `DOSEventBegin() { .. }`

Adding the `void` to the .c file fixes the error, presumably because otherwise the signatures do not match?

madebr:
I notice my IDE (=clion) also shows this warning when I hover over `DosEventEnd()`.
Let's apply this one, and in a future pr do an effort to make the code compatible with `-Wmissing-declarations`.

madebr (approved):

PR #323, opened 2023-05-04
input: use PDKeyDown3 in PDGetASCIIFromKey

### LINKED ISSUE #321: Holding down SHIFT makes key repeat go crazy quick
In places where you can enter text, such as the Driver Selection menu or Save Game menu, when you hold down the <kbd>SHIFT</kbd> key while entering text, the text is repeated very fast.

## How to reproduce

1. Start a new game
2. Click on one of the drivers
3. Remove all text using backspace
4. Hold down the <kbd>SHIFT</kbd> (with your pink)
5. Press any other letter key

## Expected result

The letter is entered, with a moderate speed

## Actual result

The complete input box is filled with the letter

madebr:
PDKeyDown3 does not modify last_key_down and last_key_down_time

`PDKeyDown3` is used in the german DOS Carmageddon executable,
but not in the GOG splatpack Windows executable.

So I'm unsure what the correct approach is here.

Fixes #321

### DIFF
--- a/src/DETHRACE/pc-win95/win95sys.c
+++ b/src/DETHRACE/pc-win95/win95sys.c
@@ -308,7 +308,7 @@ void Win32ReleaseInputDevice(void) {
 int PDGetASCIIFromKey(int pKey) {
     LOG_TRACE("(%d)", pKey);
 
-    if (PDKeyDown(KEY_SHIFT_ANY)) {
+    if (PDKeyDown3(KEY_SHIFT_ANY)) {
         return gASCII_shift_table[pKey];
     } else {
         return gASCII_table[pKey];
--- a/src/DETHRACE/pc-win95/win95sys.c
+++ b/src/DETHRACE/pc-win95/win95sys.c
@@ -308,6 +308,7 @@ void Win32ReleaseInputDevice(void) {
 int PDGetASCIIFromKey(int pKey) {
     LOG_TRACE("(%d)", pKey);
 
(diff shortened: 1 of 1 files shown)


src/DETHRACE/pc-win95/win95sys.c
```diff
@@ -308,7 +308,7 @@ void Win32ReleaseInputDevice(void) {
 int PDGetASCIIFromKey(int pKey) {
     LOG_TRACE("(%d)", pKey);
 
-    if (PDKeyDown(KEY_SHIFT_ANY)) {
+    if (PDKeyDown3(KEY_SHIFT_ANY)) {
```
dethrace-labs, line 311: lets put a comment here describing why we are deviating from the OG windows code

dethrace-labs:
Interesting! I didn't look at the german executables at all :)

I think generally our approach should be to take the best implementation parts from all the different executables to provide the 'best' experience.

As long as we comment where and why we are deviating from the default OG source, I think its fine :)

dethrace-labs (approved):

src/DETHRACE/pc-win95/win95sys.c
madebr, line 311:
```suggestion
    /* The Windows Carmageddon executable uses PDKeyDown here. The German DOS executable uses PDKeyDown3. */
    if (PDKeyDown3(KEY_SHIFT_ANY)) {
```

PR #329, opened 2023-05-18
Per-channel audio buffers

### LINKED ISSUE #281: Player car engine sound sometimes cuts out

### LINKED ISSUE #324: The horn has a noticeable clicking sound
When holding down the horn ('H' key in race), there is a continuous click/pop sound as the horn sound loops.

This does not exist in OG, and is due to miniaudio not loading the wav file correctly. `SDL_mixer` also has the same issue.

dethrace-labs:
Pulls logic from the DOS executable which implemented per-channel audio correctly. Each channel holds a reference to the playing sound instance.

In the windows build, the DirectSound implementation was flawed as the directsound buffer was attached to the sound _descriptor_, not the sound _instance_. This meant that only a single instance of a sound could be played at any time.

Also switches s3 function names from “dispose” to “release” to better match the known naming convention.

Fixes #281 
Fixes #324

### DIFF
--- a/src/S3/audio.h
+++ b/src/S3/audio.h
@@ -24,17 +24,17 @@ int S3OpenSampleDevice(void);
 void S3CloseDevices(void);
 tS3_outlet* S3CreateOutlet(int unk1, int pChannel_count);
 int S3CreateOutletChannels(tS3_outlet* outlet, int pChannel_count);
-void S3DisposeOutlet(tS3_outlet* outlet);
+void S3ReleaseOutlet(tS3_outlet* outlet);
 int S3UnbindChannels(tS3_outlet* outlet);
-void S3DisposeUnboundChannels(void);
+void S3ReleaseUnboundChannels(void);
 tS3_channel* S3AllocateChannel(tS3_outlet* outlet, int priority);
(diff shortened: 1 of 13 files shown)


src/S3/audio.h
```diff
@@ -24,17 +24,17 @@ int S3OpenSampleDevice(void);
 void S3CloseDevices(void);
 tS3_outlet* S3CreateOutlet(int unk1, int pChannel_count);
 int S3CreateOutletChannels(tS3_outlet* outlet, int pChannel_count);
-void S3DisposeOutlet(tS3_outlet* outlet);
+void S3ReleaseOutlet(tS3_outlet* outlet);
```
madebr, line 27: Asking for rec2, did you invent these names yourself?

dethrace-labs: Largely, yes. Although there are some DRS3* wrapper functions that we do know the names of in the dethrace code, so I used those where possible for names and conventions

Issue #330, opened 2023-05-20
[actions-build] Couldn't open keymap file


nistvan86:
I was about to test win64 build at https://github.com/dethrace-labs/dethrace/actions/runs/5019323827
If I enter the configuration menu and try to change to any of the layouts the game crashes, and in the log I can see.

    Failed to open "d:\tmp\dethrace/DATA/KEYMAP_0.TXT"Couldn't open Key Map file

Checking my installation those files are clearly there with content.
This works with 0.6.0.

madebr:
Can you try the executable that is built by the actions of https://github.com/dethrace-labs/dethrace/pull/331?
It won't fix the error, but should print more information about why the file could not be opened.

nistvan86:
```
Failed to open "d:\tmp\dethrace/DATA/KEYMAP_0.TXT" (Permission denied)
Couldn't open Key Map file
```

This is rather interesting as I do nothing else than overwriting the three files coming from the zip inside the game folder. I do the same restoring it with the 0.6.0 version.

madebr:
I can reproduce the error on Linux when I make `DATA/KEYMAP_0.TXT` read-only (`chmod a-w KEYMAP_*.TXT`).
Can you verify the file permissions?

nistvan86:
I see nothing out of ordinary with the files, authenticated users on the system have full access over them.

As I've said I'm using the exact same game files with the 0.6.0 and the github built version, I'm only replacing dethrace.exe, dethrace.pdb, sdl2.dll files in the game directory with either zip variant. With 0.6.0 it can load the keymap, with the other one it cannot. The game files are from the GoG installer of the game.
Actually it's enough to swap the dethrace.exe, as it still launches with the 0.6.0 dethrace.pdb and sdl2.dll files, this already shows this symptom.

I've checked and no other dethrace.exe is running on the system. Also tried adding the whole folder to Windows Defender's exclusion list but it still doesn't work. (I think it would completely block the execution if it finds it malicious, not just selective block some operations)

I'm out of ideas what this could be.

Squall-Leonhart:
Definitely an issue introduced with more recent changes, I'll start bisecting once i get the build environment setup.

Squall-Leonhart:
Heres something funny, you can change the presets just fine if you launch dethrace from ConEmu, but not from CMD prompt or powershell.

Squall-Leonhart:
[...]

PR #331, opened 2023-05-20
Little touches to windows platform code

madebr:
- Print error message when `fopen_s` fails
- Make more functions static
- Remove leading `_` from global variable (=undefined behavior)


.github/workflows/workflow.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
           cd build
           ctest --verbose
       - name: Upload Artifact
-        if: startsWith(github.ref, 'refs/tags/') || github.ref_name == 'main'
+#        if: startsWith(github.ref, 'refs/tags/') || github.ref_name == 'main'
```
madebr, line 45: I disabled this line such that the user at https://github.com/dethrace-labs/dethrace/issues/330 can download a built binary.

dethrace-labs (approved):

PR #345, opened 2023-08-07
chdir to exe path

### LINKED ISSUE #308: Dethrace fails to find gamedata when started through the Launcher (on Macos)
From discord:

> and launch it from terminal because there’s a bug where game fails to pick files otherwise

I remember reading somewhere macos sets the working directory to `/` or something similar. Or it adds some arguments? Idk, I don't use it.

I think we should detect this case, and use [`_NSGetExecutablePath`](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/dyld.3.html) to change the working directory.

dethrace-labs:
This patch uses `DETHRACE_ROOT_DIR` if defined, otherwise, uses the executable path to set the root directory

Fixes #308

### DIFF
--- a/src/harness/include/harness/os.h
+++ b/src/harness/include/harness/os.h
@@ -30,8 +30,8 @@ FILE* OS_fopen(const char* pathname, const char* mode);
 
 size_t OS_ConsoleReadPassword(char* pBuffer, size_t pBufferLen);
 
-char* OS_Dirname(char* path);
+char* OS_Dirname(const char* path);
 
-char* OS_Basename(char* path);
+char* OS_Basename(const char* path);
 
 #endif
--- a/src/harness/include/harness/os.h
+++ b/src/harness/include/harness/os.h
@@ -30,4 +30,8 @@ FILE* OS_fopen(const char* pathname, const char* mode);
 
 size_t OS_ConsoleReadPassword(char* pBuffer, size_t pBufferLen);
 
+char* OS_Dirname(char* path);
+
+char* OS_Basename(char* path);
+
 #endif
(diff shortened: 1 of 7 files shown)


src/harness/os/windows.c
```diff
@@ -169,3 +169,15 @@ size_t OS_ConsoleReadPassword(char* pBuffer, size_t pBufferLen) {
     fgets(pBuffer, pBufferLen, stdin);
     return strlen(pBuffer);
 }
+
+char* dir[_MAX_DIR];
+char* fname[_MAX_FNAME];
```
madebr, line 174:
```suggestion
static char* dir[_MAX_DIR];
static char* fname[_MAX_FNAME];
```

src/harness/os/linux.c
```diff
@@ -301,3 +301,11 @@ size_t OS_ConsoleReadPassword(char* pBuffer, size_t pBufferLen) {
...
     return len;
 }
+
+char* OS_Dirname(char* path) {
+    return dirname(path);
+}
+
+char* OS_Basename(char* path) {
+    return basename(path);
+}
```
madebr, line 311: These functions may modify their argument.
Perhaps also use a static buffer, as it does on Windows?

src/harness/include/harness/os.h
```diff
@@ -30,4 +30,8 @@ FILE* OS_fopen(const char* pathname, const char* mode);
 
 size_t OS_ConsoleReadPassword(char* pBuffer, size_t pBufferLen);
 
+char* OS_Dirname(char* path);
+
+char* OS_Basename(char* path);
```
madebr, line 35: With my other suggestions about static storage, can the argument be `const char *`?

Issue #352, opened 2023-12-18
Improve support for the Brazilian Portuguese version


leomontenegro6:
A few months ago, I found out this cool reverse engineering project of Carmageddon. Glad it worked fine not only in PC, but also in some linux-based portable devices, such as Anbernic ones.

The original game works fine, but out of my curiosity, I tested if dethrace wouldn't, for some reason, support the [Brazilian Portuguese version](https://archive.org/details/Carma) of this game, that features lots of texts and graphical changes, and even voice dubbing. It worked to some extend, but lots of menus were bugged. It worked in the Windows PC version, but crashed in Linux when I tried to run in my device with [the binary compiled by PortMaster](https://portmaster.games/detail.html?name=carmageddon). Not sure if it's something from the compilation, or if it's really an issue with any dethrace linux versions.

Is there a chance we could improve dethrace to fully support this version?

madebr:
I just tested dethrace of current master with the Portuguese game data, and some menus indeed cause a segmentation fault (e.g. the Race Selection Menu).

About the menus, if you're talking about the text disappearing and re-appearing when hovering your mouse, I think OG also has this.
See [this video](https://youtu.be/5J82nMjJstM?t=137): the text also disappears there.
This behavior also occurs with German game data.

The options menu also looks bugged.

madebr:
With #353, Portuguese dethrace does not crash anymore in the menus.

leomontenegro6:
Oh, that's pretty cool. I'll let PortMaster guys know about this update, and see if we can recompile it. Not sure if we're gonna do that with master, or after a new stable version is released with these changes.

About the YT video above, is there a difference between grabbing original game and merging the translated files with it, or getting all the data directly from a PT-BR ISO image? Is the result going to be the same, or did the author make some changes to accommodate the files with the original game?

PR #353, opened 2023-12-19
Fixes for Portuguese Carmageddon

madebr:
The Portuguese Carmageddon segfaulted because its character set caused signed characters to access out-of-bounds memory.

I also added a few fixes for warnings/errors seen in rec2.

Fixes https://github.com/dethrace-labs/dethrace/issues/352

### DIFF
--- a/src/BRSRC13/CORE/FW/datafile.c
+++ b/src/BRSRC13/CORE/FW/datafile.c
@@ -308,15 +308,18 @@ int TextReadLine(br_datafile* df, char** ident, char** data) {
     LOG_TRACE9("(%p, %p, %p)", df, ident, data);
 
     while (1) {
+        int a;
         cp = BrScratchString();
-        int a = BrFileGetLine(cp, 256, df->h);
+        a = BrFileGetLine(cp, 256, df->h);
+        (void)a;
         if (BrFileEof(df->h) != 0) {
             return 0;
         }
         for (; (*cp == ' ') || (*cp == '\t'); cp++) {
         }
-        if (*cp != '\0')
+        if (*cp != '\0') {
             break;
+        }
     }
     *ident = cp;
     while ((*cp != ' ') && (*cp != '\t') && (*cp != '\0')) {
--- a/src/DETHRACE/common/options.c
+++ b/src/DETHRACE/common/options.c
@@ -468,7 +468,7 @@ void PlotAGraphBox__options(int pIndex, int pColour_value) {
         gThe_interface_spec__options->mouse_areas[2].left[gGraf_data_index] - 6,
         gRadio_bastards__options[pIndex].top - 3,
         gThe_interface_spec__options->mouse_areas[2].right[gGraf_data_index] + 3,
-        gRadio_bastards__options[pIndex].top + gFonts[12].height + (TranslationMode() ? 2 : 0), pColour_value);
+        gRadio_bastards__options[pIndex].top + gFonts[12].height + 2 - (TranslationMode() ? 2 : 0), pColour_value);
 }
 
 // IDA: void __usercall DrawAGraphBox(int pIndex@<EAX>)
@@ -672,6 +672,9 @@ void StripControls(unsigned char* pStr) {
         if (pStr[i] < ' ') {
             memmove(&pStr[i], &pStr[i + 1], (len - i) * sizeof(char));
             len--;
+#ifdef DETHRACE_FIX_BUGS
+            i--;
+#endif
         }
     }
 }
--- a/src/BRSRC13/CORE/MATH/matrix4.c
+++ b/src/BRSRC13/CORE/MATH/matrix4.c
@@ -114,7 +114,7 @@ br_scalar BrMatrix4Inverse(br_matrix4* A, br_matrix4* B) {
     if (fabs(det) < BR_SCALAR_EPSILON * 2)
         return 0;
 
-    idet = 1.0 / det;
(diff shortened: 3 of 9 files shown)


PierreMarieBaty:
A possibly related problem: the French version of Carmageddon has text file translations in which accentuated characters over ASCII value 127 are used. This is clearly a mistake from the translator as no corresponding glyphs can represent these characters, but it makes an assertion fail in src/DETHRACE/common/utility.c in the GetALineWithNoPossibleService() function:
```
    } while (!isalnum(s[0]) // <-- assertion failure
        && s[0] != '-'
        && s[0] != '.'
        && s[0] != '!'
        && s[0] != '&'
        && s[0] != '('
        && s[0] != '\''
        && s[0] != '\"'
        && s[0] >= 0);
```
When ignoring the assertion, the game runs, but these characters are displayed as blank in the game messages (note that this is consistent with the DOS version).

madebr:
Using current dethrace master, I don't see glyph related issues with dethrace, using [this French game data from archive.org](https://archive.org/details/carmageddon-pc-france).

(I see a crash on Linux because the audio backend cannot handle audio files with wrong cases)

madebr:
The last commit fixes the crash due to some French audio files having a wrong casing.

PierreMarieBaty:
Don't you see missing characters in the menu dialogs where accentuated 
characters should be used ?

https://freeimage.host/i/J7zfjYQ

On my version, accentuated characters are missed out, and the 
"unbreakable space" used by the translator before colons (such as in 
"xxx<here>: yyy" yields a black block. Another character provokes a 
carriage return and sometimes a crash.

If you don't see such things, then I guess there were several releases 
of the game in french and I have an early one...

madebr:
Oh yes, now I see it. I was expecting a crash and did not read all texts.
Muscle memory knows what buttons to press :)

So yes, this pr will fix these messages.

Screenshot from 2024-01-06 14-41-11

madebr:
I was wrong, the `à` in "Retourner à la vie de tous les jours, péniblement calme?" is still missing.

madebr:
`DATA/TRNSLATE` contains this:
```
QUT3STIL,164,34,1,C,L,QUITTER
QUT3STIL,164,80,1,C,L,RETOURNER  LA VIE 
QUT3STIL,164,91,1,C,L,DE TOUS LES JOURS, 
QUT3STIL,164,102,1,C,L,P<C9>NIBLEMENT CALME?
QUT3STIL,116,130,2,C,L,ANNULER
QUT3STIL,213,130,2,C,L,O.K.
```
(The format is `<FLICNAME>,<x>,<y>,<font>,<just>`)

So apart from explicitly doing a `strcmp`, I don't think we can't fix this.
The translators forgot to add the "À".

When I add it manually (hexcode `0xc0`), it works:
image

PierreMarieBaty:
Good point. But there's more to it than just this occurence.

See for example DATA/RACES.TXT line 102 (fourth race)

situation<A0>: mine du
canyon du diable

This will display a black block instead of the non-breakable space.

And other characters in other places... As soon as the character gets 
over 0x80 it's likely to not be rendered correctly.

madebr:
I see the missing whitespace before the colon for the 4th race.
I think it's not too unreasonable for dethrace to require correct input data (as long as it does not crash).
If you want these fixed, I think that works need to be done in a patch set, served by moddb.com or rr2000.

PierreMarieBaty:
TBH I don’t care as I already patched all my French data files. I was just reporting the problem as it’s something I’ve come across and I thought other people would encounter too.

Anyway, defaulting a character to a space (or to nothing) and emitting a console warning when it’s not among the drawable glyphs of the selected font is a quick fix, and would improve the robustness of the app. That’s the way I’d choose in my own code.

madebr:
Printing a warning is not unreasonable, but we need to be careful to not spam the console.
In some places, carmageddon will draw text every frame.

src/BRSRC13/CORE/FW/datafile.c
```diff
@@ -308,15 +308,18 @@ int TextReadLine(br_datafile* df, char** ident, char** data) {
     LOG_TRACE9("(%p, %p, %p)", df, ident, data);
 
     while (1) {
+        int a;
         cp = BrScratchString();
-        int a = BrFileGetLine(cp, 256, df->h);
+        a = BrFileGetLine(cp, 256, df->h);
+        (void)a;
```
dethrace-labs, line 314: what is this doing?

src/DETHRACE/common/options.c
```diff
@@ -672,6 +672,9 @@ void StripControls(unsigned char* pStr) {
         if (pStr[i] < ' ') {
             memmove(&pStr[i], &pStr[i + 1], (len - i) * sizeof(char));
             len--;
+#ifdef DETHRACE_FIX_BUGS
+            i--;
```
dethrace-labs, line 676: what behavior does this fix?

dethrace-labs (approved):

src/BRSRC13/CORE/FW/datafile.c
madebr: I silences a warning about setting a variable, but not using it.
It's probably better to just remove the `a = ` thing.

src/DETHRACE/common/options.c
dethrace-labs, line 676:
```suggestion
            // correctly handle stripping multiple control characters
            i--;
```

PR #362, opened 2024-05-26
ci: use libsdl-org/setup for SDL + package with CPack

madebr:
- Fix building dethrace from a source archive
- Disable the test that accessed memory after it had been freed
- use libsdl-org/setup-sdl for configuring SDL on ci
- package release binaries with CPack

Normal ci output: https://github.com/madebr/dethrace/actions/runs/9244495833
Release ci output: https://github.com/madebr/dethrace/actions/runs/9245171693
Dummy release: https://github.com/madebr/dethrace/releases/tag/v0.0.69


dethrace-labs:
Looks nice. The windows packages are missing SDL2.dll though, which we should include. Also, for the macos package, I think having the LICENSE file in the root of the archive like windows rather than in a `/share/..` directory is more expected.

In the future we'll need to move the macos target to build a proper .app package with icon and SDL2 libraries packaged. I was looking at https://github.com/diasurgical/devilutionX/releases/tag/1.5.2, they had a lot of targets and it would would be great to build our infrastructure in a way that we could end up with something similar :)

madebr:
Made it install to the root for all packages.
About creating a `.app` package, I think we need to do `set_property(TARGET dethrace PROPERTY MACOSX_BUNDLE TRUE)`.
I think you can then add `libSDL2.dylib` to it (and configure dethrace's runpath such that it can actually find SDL2)

dethrace-labs (approved):

Issue #350, opened 2023-08-11
Render glitches while in in-car view


Gagert:
Using build 4200cfa on Windows 11, there seems to be various rendering glitches while in in-car view. You can see colored trails of the cars passing by at the start of the match, while at the middle of the image the skybox & environments at the center of the view is glitching & duplicating.

The rear mirror is currently black (known issue I suppose); I don't know if the game has to be rendered twice to show the rear mirror view, or what kind of tricks the game originally used to achieve this effect.

![carma in-car bug](https://github.com/dethrace-labs/dethrace/assets/8365446/0e02ae47-c7af-4350-a2a5-8be724bfc0b8)

dethrace-labs:
Fixed by https://github.com/dethrace-labs/dethrace/pull/363

Issue #209, opened 2022-10-09
Damage gallery not displaying cursor around selected car


rafalp:
Original game displays bounding box around selected wreck, as is illustrated on Carma's wiki:

![Wiki](https://static.wikia.nocookie.net/carmageddon/images/d/d5/Menu-C1-damagegallery.gif/revision/latest?cb=20090829144605)

dethrace-labs:
Fixed by #363

PR #363, opened 2024-06-12
Software renderer

### LINKED ISSUE #354: Rear view mirror in cockpit mode is not implemented
The rear view mirror is [currently disabled](https://github.com/dethrace-labs/dethrace/blob/8051ca2b6d3e060321407c69a65f92548bfe68e0/src/DETHRACE/common/graphics.c#L1693-L1717).
Uncommenting it is not enough to get it working.

dethrace-labs:
- [BRender-v1.3.2](https://github.com/dethrace-labs/BRender-v1.3.2) pulled in via submodule
- Multiple FIXME/HACKs resolved by using the 'real' BRender instead of the reverse engineered one
- Removed all opengl bits and minified a decent amount of harness
- Rearview mirror enabled
- Adds `linux-arm64` build
Screenshot 2024-05-30 at 9 55 02 PM
Screenshot 2024-05-27 at 8 04 16 PM

Fixes #354, #350, #209,


.github/workflows/workflow.yaml
```diff
@@ -16,6 +16,7 @@ jobs:
       matrix:
         platform:
           - { name: 'Linux',    arch: 'x64',        os: ubuntu-latest, werror: true }
+          - { name: 'Linux',    arch: 'arm64',      os: ubuntu-latest, werror: true }
```
madebr, line 19: You need to install [gcc-aarch64-linux-gnu](https://packages.ubuntu.com/jammy/gcc-aarch64-linux-gnu) (and [g++-aarch64-linux-gnu](https://packages.ubuntu.com/jammy/g++-aarch64-linux-gnu)), and also use a CMake toolchain file.

The CMake toolchain file can be as simple as:
```cmake
set(CMAKE_SYSTEM_NAME "Linux")
set(CMAKE_SYSTEM_PROCESSOR "arm64")
set(CMAKE_C_COMPILER "aarch64-linux-gnu-gcc")
set(CMAKE_CXX_COMPILER "aarch64-linux-gnu-g++")
```

madebr: I (ab)used my rights, and fixed it in d62cd4fd09d64cb82c58c2405888b25f4442e838

madebr:
I get a build error in `src\harness\os\windows.c` when targeting arm64 on Windows.
See [this log](https://github.com/madebr/dethrace/actions/runs/9483856579/job/26131972785#step:9:672).

madebr:
Badly damaging (until it burns) and repairing the yellow eagle causes corruption of the repaired parts.
Screenshot from 2024-06-12 20-04-31

.github/workflows/workflow.yaml
dethrace-labs: ah thanks. Oops, I expected that an arm64 ubuntu box would run the action, but I realize now that arm runners are only in private beta. Thanks for adding the compilation!

dethrace-labs:
thanks! That reminds me I saw something similar on Val Hella. I'll take a look

dethrace-labs:
Fixed the incorrect rendering on the yellow eagle. The issue was caused by an [incorrect line in the arbitrary width lit triangle path](https://github.com/dethrace-labs/BRender-v1.3.2/commit/9b6c6d99822f37ad226043b305669fa835822ae0). The reason it only appears in yellow eagle is the other colors use power-of-two textures which go through a different path.

Screenshot 2024-06-13 at 9 51 14 PM

madebr (approved):
I'm testing the network code on top of this branch and can see no regressions.
The game runs fine in Debug mode.
So let's merge this and fix bugs as we encounter them.

madebr:
About the BRender-v1.3.2 fork, are we supposed to open issues here?
Through rec2, I think I found a few bugs/typos in the brender code on which I'd like your feedback.

dethrace-labs:
Yep, I think that makes sense. BRender-specific issues should go in that repo

xcom169:
Which hardware renderer can we expect along with the cpu-renderer? Glide?

dethrace-labs:
As of this commit, we will only have the software renderer. Then, I'll add back the hardware renderer, implementing the original 3dfx code paths

PR #365, opened 2024-06-13
Fix windows chdir behavior

dethrace-labs:
`argv[0]` was expected to include the full path to the dethrace executable. On windows, this is not true, and only includes the filename. 

As a fix, we should only try to `chdir` if `root_dir` is not empty or a single character.

### DIFF
--- a/src/harness/harness.c
+++ b/src/harness/harness.c
@@ -181,7 +181,8 @@ void Harness_Init(int* argc, char* argv[]) {
     } else {
         root_dir = OS_Dirname(argv[0]);
     }
-    if (root_dir != NULL && strlen(root_dir) > 1) {
+    // if root_dir is empty or ".", no need to chdir
+    if (root_dir != NULL && root_dir != '\0' && root_dir != '.') {
         printf("Using root directory: %s\n", root_dir);
         result = chdir(root_dir);
         if (result != 0) {
--- a/src/harness/harness.c
+++ b/src/harness/harness.c
@@ -181,10 +181,12 @@ void Harness_Init(int* argc, char* argv[]) {
     } else {
         root_dir = OS_Dirname(argv[0]);
     }
-    printf("Using root directory: %s\n", root_dir);
-    result = chdir(root_dir);
-    if (result != 0) {
-        LOG_PANIC("Failed to chdir. Error is %s", strerror(errno));
+    if (root_dir != NULL && strlen(root_dir) > 1) {
+        printf("Using root directory: %s\n", root_dir);
+        result = chdir(root_dir);
+        if (result != 0) {
+            LOG_PANIC("Failed to chdir. Error is %s", strerror(errno));
+        }
     }
 
     if (harness_game_info.mode == eGame_none) {
--- a/src/harness/harness.c
+++ b/src/harness/harness.c
@@ -181,7 +181,7 @@ void Harness_Init(int* argc, char* argv[]) {
     } else {
         root_dir = OS_Dirname(argv[0]);
     }
-    // if root_dir is empty or ".", no need to chdir
+    // if root_dir is null or empty, no need to chdir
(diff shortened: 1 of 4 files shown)


src/harness/harness.c
```diff
@@ -181,10 +181,12 @@ void Harness_Init(int* argc, char* argv[]) {
     } else {
         root_dir = OS_Dirname(argv[0]);
     }
-    printf("Using root directory: %s\n", root_dir);
-    result = chdir(root_dir);
-    if (result != 0) {
-        LOG_PANIC("Failed to chdir. Error is %s", strerror(errno));
+    if (root_dir != NULL && strlen(root_dir) > 1) {
```
madebr, line 184: On discord, you said `"."` worked.
What about only ignoring an empty `root_dir`?
```suggestion
    if (root_dir != NULL && root_dir[0] != '\0') {
```

dethrace-labs: My thinking was if the value is `.`, we don't need to call chdir since it is a no-op. I suppose I could do a `strcmp(root_dir, '.')` to make that more obvious

src/harness/renderers/gl/gl_renderer.c
```diff
@@ -220,6 +220,10 @@ void GLRenderer_Init(int pRender_width, int pRender_height) {
     LOG_INFO("OpenGL version string: %s", glGetString(GL_VERSION));
     LOG_INFO("OpenGL shading language version string: %s", glGetString(GL_SHADING_LANGUAGE_VERSION));
 
+    if (glGetString(GL_SHADING_LANGUAGE_VERSION) == NULL) {
```
dethrace-labs, line 223: this resolves a null pointer error inside `glCreateProgram` later on inside my windows VM which doesn't expose gl shaders

src/harness/harness.c
madebr: It's a no-op, but I don't think we don't care. `chdir`'s implementation should be able to handle it.
The current change would disallow one-letter directories.
An unlikely directory structure, but possible.

src/harness/harness.c
```diff
@@ -181,10 +181,13 @@ void Harness_Init(int* argc, char* argv[]) {
     } else {
         root_dir = OS_Dirname(argv[0]);
     }
-    printf("Using root directory: %s\n", root_dir);
-    result = chdir(root_dir);
-    if (result != 0) {
-        LOG_PANIC("Failed to chdir. Error is %s", strerror(errno));
+    // if root_dir is empty or ".", no need to chdir
+    if (root_dir != NULL && root_dir != '\0' && root_dir != '.') {
```
madebr, line 185: This would disallow data in a (hidden) dotdirectory, e.g. `.carmageddon1`.
Unlikely, but still possible :)

dethrace-labs: haha yep, ok, I'm convinced :)

src/harness/renderers/gl/gl_renderer.c
madebr: What OpenGL version does the vm expose?

But you're probably correct to test capabilities instead of requiring a minimum OpenGL version.

src/harness/harness.c
```diff
@@ -181,10 +181,13 @@ void Harness_Init(int* argc, char* argv[]) {
     } else {
         root_dir = OS_Dirname(argv[0]);
     }
-    printf("Using root directory: %s\n", root_dir);
-    result = chdir(root_dir);
-    if (result != 0) {
-        LOG_PANIC("Failed to chdir. Error is %s", strerror(errno));
+    // if root_dir is empty or ".", no need to chdir
```
madebr, line 184:
```suggestion
    // if root_dir is null or empty, no need to chdir
```

madebr (approved):
With my last comment, LGTM!

dethrace-labs:
thanks for the spotting! I must be tired today :)

PR #371, opened 2024-06-17
Fix Windows arm64 build

madebr:
I disabled the ci job, but when enabled it would look as follows:
https://github.com/madebr/dethrace/actions/runs/9550571423

Also, it's completely untested


ArminiusTux (approved):
With these changes `dethrace` now successfully compiles und runs under Windows on ARM (Win 11 v23H2).

Tested with MSVC (VS v17.10.3) & Clang v18.1.6.

ArminiusTux:
Thank you @madebr for this fix!
How about a neat icon for this game - maybe @Toshiba-3 or @Memorix101 could help?

madebr:
> Thank you @madebr for this fix! How about a neat icon for this game - maybe @Toshiba-3 or @Memorix101 could help?

Great to hear!
The arm64 incompatibility was in the error reporting code, which you hopefully never see.
You'd need to purposefully add a crashing bug to the game to verify my fixes are correct.
If they are working, the code should print a stacktrace of the crash location.

@dethrace-labs 
Perhaps we should use dbghelp instead: [CaptureStackBackTrace](https://learn.microsoft.com/en-us/windows/win32/debug/capturestackbacktrace), [Syminitialize](https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-syminitialize), [SymFromAddr](https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symfromaddr) and [SymGetLineFromAddr](https://learn.microsoft.com/en-us/windows/win32/api/dbghelp/nf-dbghelp-symgetlinefromaddr) are way simpler and not architecture dependent.
They are used in the SDL tests to track memory allocations (and detect leaks).

dethrace-labs:
@madebr agreed - when I had a quick look on this issue I thought CaptureStackBackTrace et all would be much simpler.

@ArminiusTux for the icon I’m working on building a proper app package for MacOS and as part of that will include the windows icon config

madebr:
I looked into stacktrace generation past weekend, and I don't think `CaptureStackBackTrace` is useful here.
That function captures the stack of the current thread stack where in an error state, we want to print the stack of another failing thread stack.
`CaptureStackBackTrace` is useful for assert-like functions to show where the assertion failed, and how you got there.

I used these somewhat complicated functions to print stacktraces of crashed child processes on ci.
https://github.com/libsdl-org/SDL/blob/main/test/win32/sdlprocdump.c

src/harness/os/windows.c
```diff
@@ -142,15 +170,25 @@ static LONG WINAPI windows_exception_handler(EXCEPTION_POINTERS* ExceptionInfo)
     if (EXCEPTION_STACK_OVERFLOW != ExceptionInfo->ExceptionRecord->ExceptionCode) {
         print_stacktrace(ExceptionInfo->ContextRecord);
     } else {
-        addr2line(windows_program_name, (void*)ExceptionInfo->ContextRecord->Eip);
+#if defined(DETHRACE_CPU_X86) || defined(DETHRACE_CPU_X64)
+        void *addr = (void*)ExceptionInfo->ContextRecord->Eip;
+#elif defined(DETHRACE_CPU_ARM64)
+        void *addr = (void*)ExceptionInfo->ContextRecord->Pc;
+#endif
+        addr2line(windows_program_name, addr);
```
dethrace-labs, line 178: Can we rely on having an `addr2line` executable on windows? We shouldn't use `SymGetLineFromAddr64` or something instead?

madebr: Nope. That's an error.

madebr:
My latest changes will also print a stacktrace-like on dethrace where dbghelp is not 100% suported, and when `addr2line.exe` does not exist:
```
Dethrace version: v0.7.1-13-g0ad96de2
Using root directory: \home\maarten\projects\dethrace\cmake-build-debug-mingw64\
Game mode: Carmageddon
Error: EXCEPTION_INT_DIVIDE_BY_ZERO
0: Z:\home\maarten\projects\dethrace\cmake-build-debug-mingw64\dethrace.exe 0x40126f20
1: Z:\home\maarten\projects\dethrace\cmake-build-debug-mingw64\dethrace.exe 0x400012ee
2: Z:\home\maarten\projects\dethrace\cmake-build-debug-mingw64\dethrace.exe 0x400013e6
3: C:\windows\system32\kernel32.dll 0xffa88c79
4: C:\windows\system32\ntdll.dll 0xffc86e8f
```

dethrace-labs (approved):

PR #381, opened 2024-06-27
Adds proper MacOS bundle and icons for Windows builds

dethrace-labs:
- MacOS binary bundled with the SDL dependency as a `.app` , distributed inside a disk image.
- Windows binary contains an icon

Windows:
Screenshot 2024-07-03 at 1 55 46 PM

MacOS:
Screenshot 2024-07-03 at 1 56 47 PM


src/DETHRACE/CMakeLists.txt
```diff
@@ -195,8 +219,16 @@ endif()
 
 if (DETHRACE_INSTALL)
     install(TARGETS dethrace
-        RUNTIME DESTINATION "."
+        BUNDLE DESTINATION . COMPONENT Runtime
+        RUNTIME DESTINATION bin COMPONENT Runtime
```
madebr, line 223:
```suggestion
        RUNTIME DESTINATION "${CMAKE_INSTALL_BINDIR}" COMPONENT Runtime
```
And do `include(GNUInstallDirs)` somewhere before this point.

madebr: This also means the windows zip will have `dethrace.exe` in a bin directory.
So perhaps we want to configure with `-DCMAKE_INSTALL_BINDIR=.` on Windows,
or do the following before `include(GNUInstallDirs)`:
```cmake
if(WIN32)
  set(CMAKE_INSTALL_BINDIR "." CACHE PATH "Binary installation directory")
endif()
include(GNUInstallDirs)
```

Also, make sure to install the extra runtime dlls (SDL2.dll) and the pdb (SDL2.pdb) to the same directory in [the lines below](https://github.com/dethrace-labs/dethrace/blob/673244bb8c19cdd3041bfd69e0891c504f58757a/src/DETHRACE/CMakeLists.txt#L196-L212)

dethrace-labs: How about set RUNTIME DESTINATION “.” as it was before maybe?

madebr: That's fine!
If somebody ever packages this, he can move it himself :)

src/DETHRACE/CMakeLists.txt
```diff
@@ -195,8 +202,40 @@ endif()
 
 if (DETHRACE_INSTALL)
     install(TARGETS dethrace
-        RUNTIME DESTINATION "."
+        BUNDLE DESTINATION . COMPONENT Runtime
+        RUNTIME DESTINATION . COMPONENT Runtime
     )
+
+    if (APPLE)
```
madebr, line 209: Can you fix the indentation and make sure all cmake commands are lower case?

src/DETHRACE/CMakeLists.txt
```diff
@@ -195,8 +200,38 @@ endif()
...
+            MACOSX_BUNDLE_BUNDLE_VERSION ${DETHRACE_VERSION}
+            MACOSX_BUNDLE_SHORT_VERSION_STRING ${DETHRACE_VERSION}
+            MACOSX_BUNDLE_ICON_FILE dethrace.icns
+        )
+        set_source_files_properties(${CMAKE_SOURCE_DIR}/packaging/macos/dethrace.icns PROPERTIES
+            MACOSX_PACKAGE_LOCATION "Resources"
+        )
+        target_sources(dethrace_obj PRIVATE
+            ${CMAKE_SOURCE_DIR}/packaging/macos/dethrace.icns
+        )
```
madebr, line 222: Do we need an option to build DethRace on macos *not* as a bundle, as an ordinary cli executable?

dethrace-labs: without `-DDETHRACE_INSTALL`, and ordinary executable is created

madebr: True :)
That seems good enough for me.

PR #384, opened 2024-07-05
Fix some Address Sanitizer errors

dethrace-labs:
Running with address sanitizer raised a couple of errors:

- `CollCheck` made some assumptions that the colliding object is a `tCar_spec`, when it could be a `tNon_car_spec`. 
- `CreateSplash` makes similar assumptions. 

A fairly reliable repro is to collide with the wrecked car next to the water in Maim Street and knock it into the water

### DIFF
--- a/src/DETHRACE/CMakeLists.txt
+++ b/src/DETHRACE/CMakeLists.txt
@@ -10,11 +10,12 @@ target_include_directories(dethrace_obj
         pd
 )
 
-# add_compile_options(-fsanitize=address)
-# add_link_options(-fsanitize=address)
+if (DETHRACE_ASAN)
+    target_compile_options(dethrace_obj PUBLIC -fsanitize=address)
+    target_link_options(dethrace_obj PUBLIC -fsanitize=address)
+endif()
 
 target_link_libraries(dethrace_obj PUBLIC SDL2::SDL2 smackw32 harness BRender::Full BRender::DDI s3)
-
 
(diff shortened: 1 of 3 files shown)


src/DETHRACE/CMakeLists.txt
```diff
@@ -10,12 +10,13 @@ target_include_directories(dethrace_obj
         pd
 )
 
-# add_compile_options(-fsanitize=address)
-# add_link_options(-fsanitize=address)
+if (DETHRACE_ASAN)
+    target_compile_options(dethrace_obj PUBLIC -fsanitize=address)
+    target_link_options(dethrace_obj PUBLIC -fsanitize=address)
+endif()
```
madebr, line 16: Perhaps we should do this in the top `CMakeList.txt`, and do the following instead?
```
target_compile_options(-fsanitize=address)
target_link_options(-fsanitize=addres)
```
Then we are sure everything is built with address sanitizer (including brender and 3rd party deps).
Or don't we want to build those with sanitizer support?

dethrace-labs: My intention was to only target `dethrace`. BRender already has its own sanitizer config and I didn't want to wade into the 3rd party deps!

madebr: That makes sense :)

madebr (approved):

PR #393, opened 2024-07-12
Original music

### LINKED ISSUE #278: No Music
Is the music working on dethrace? Otherwise I made a temporary solution, making a sh file with this code:
`playsound $HOME/.local/share/Steam/steamapps/common/Carmageddon1/CARMA/MUSIC/* --loop | $HOME/.local/share/Steam/steamapps/common/Carmageddon1/CARMA/dethrace -sound 4 -hires`

dethrace-labs:
- Adds support for music as in the GOG release. That is, `MUSIC/Track0[1-8].ogg` files. If they exist, the original CD audio functions are redirected to play the ogg files.
- Adds `stb_vorbis` to enable vorbis support in miniaudio
- Switches `miniaudio` to use the default header-only implementation (to make it work with vorbis)
- Fixes bug in sound options where the dials are both initially set to the music volume

Fixes #278


src/DETHRACE/common/sound.c
```diff
@@ -681,19 +681,59 @@ int GetIndexFromOutlet(tS3_outlet_ptr pOutlet) {
...
+                S3StopOutletSound(gMusic_outlet);
+                if (gSound_enabled) {
+                    if (gProgram_state.cockpit_on && gProgram_state.cockpit_image_index >= 0) {
+                        S3Service(1, 0);
+                    } else {
+                        S3Service(0, 0);
+                    }
+                    if (pCDA_id == 9999) {
+                        do {
+                            pCDA_id = gRandom_CDA_tunes[IRandomBetween(0, 7)];
```
madebr, line 697: dethrace's `IRandomBetween(0, 7)` returns variable in the range `[0, 7]` (inclusive 0 and 7).
However, `gRandom_CDA_tunes` has length 7, with maximum index 6.

The `MUSIC` folder of my gog copy has 8 tracks (+ a 4s silence track) so I think changing `gRandom_CDA_tunes` to `int[8]` makes sense.

madebr: The 7th item on my system is currently 0, which causes a stack overflow in `S3GetDescriptorByID(0)` -> `S3GetDescriptorByID(0)` -> `S3GetDescriptorByID(0)` -> ...

src/DETHRACE/common/sound.c
```diff
@@ -16,13 +16,13 @@
...
 int gVirgin_pass = 1;
 int gOld_sound_detail_level = -1;
-int gLast_tune;
-int gRandom_MIDI_tunes[3];
-int gRandom_Rockin_MIDI_tunes[3];
-int gRandom_CDA_tunes[7];
+int gLast_tune = -1;
+int gRandom_MIDI_tunes[3] = { 9500, 9501, 9502 };
+int gRandom_Rockin_MIDI_tunes[3] = { 9500, 9501, 9502 };
+int gRandom_CDA_tunes[7] = { 9600, 9601, 9602, 9603, 9604, 9605, 9606 };
```
madebr, line 25: See https://github.com/dethrace-labs/dethrace/pull/393/files#r1676807636
A similar change is needed in `sound.h`.

```suggestion
int gRandom_CDA_tunes[8] = { 9600, 9601, 9602, 9603, 9604, 9605, 9606, 9607 }; /* dethrace: Changed to size 8 */
```

src/DETHRACE/common/sound.c
dethrace-labs: Appreciate your second pair of eyes on this one! I originally made the change to `[8]`, based on there being 8 tracks in `SOUND.TXT`, and seeing `9607` in IDA at the end of the array. When I saw the last track was silence, I figured it was deliberately left off. 

I didn't think about the `IRandomBetween(0, 7)` though! :) 

I wonder if the original code was declared as 7 but initialized with 8 items like ```int gRandom_CDA_tunes[7] = { 9600, 9601, 9602, 9603, 9604, 9605, 9606, 9607 };``` maybe

madebr (approved):
I'm not sure `assert` is the right approach in `S3GetDescriptorByID` because it is removed when building a Release build.
But LGTM!

dethrace-labs:
> I'm not sure `assert` is the right approach in `S3GetDescriptorByID` because it is removed when building a Release build. But LGTM!

That was my sort of my intention. The check is not present in the original code, but the assert could be helpful if we are working on the audio code in the future and break something

PR #400, opened 2024-07-21
Fix 399: Game does not obey sound effects volume level set in sound settings.

### LINKED ISSUE #399: Game does not obey sound effects volume level set in sound settings.
**Version**
Code: v0.7.1-19-gec04a9c
Platform: Windows

**Symptoms**
If the game is started with `--sound-options` the volume level for effects does nothing. This applies to in-game spatial effects as well as menu "tick" when user navigates through options.

**Code Analysis**
The issue is observed in game running on Windows platform. However, all the platforms seem to be affected.

1. The "tick" in all menus

It's due to incorrect order of calls inside `S3PlaySample()` function.

The following change addresses the issue:
```diff
int S3PlaySample(tS3_channel* chan) {
    tS3_sample* sound_data;

    if (chan->type_struct_sample == NULL) {
        return 0;
    }

-   S3SyncSampleVolumeAndPan(chan);
-   S3SyncSampleRate(chan);

    sound_data = (tS3_sample*)chan->descriptor->sound_data;
    if (AudioBackend_PlaySample(chan->type_struct_sample,
            sound_data->channels,
            sound_data->dataptr,
            sound_data->size,
            sound_data->rate,
            chan->repetitions == 0)
        != eAB_success) {
        return 0;
    }

+   S3SyncSampleVolumeAndPan(chan);
+   S3SyncSampleRate(chan);
```

2. In-game spatial effects

It's due to incorrect transformation of volume specified in "Carmageddon decibels" (which are linear) into linear range <0.0f..1.0f>. Miniaudio `ma_volume_db_to_linear()` function expects value expressed in "real decibels" which are exponential. Using it for "Carmageddon decibels" produce either too loud or too quiet sound.

The best solution I have found is to recalculate "Carmageddon decibels" into <0.0f..1.0f> range explicitly

```diff
tAudioBackend_error_code AudioBackend_SetVolume(void* type_struct_sample, int volume_db) {
    tMiniaudio_sample* miniaudio;
    float linear_volume;

    miniaudio = (tMiniaudio_sample*)type_struct_sample;
    assert(miniaudio != NULL);

-   // convert from directsound -10000 - 0 decibel volume scale
-   linear_volume = ma_volume_db_to_linear(volume_db / 100.0f);
+   // convert from Carmageddon "decibel" scale to linear <0.0f..1.0f>
+   linear_volume = (volume_db + 350.0f) / -5.0f;
+   if (linear_volume == 0.0f) {
+       linear_volume = 1.0f;
+   }
+   linear_volume = 1.0f / linear_volume;

    ma_sound_set_volume(&miniaudio->sound, linear_volume);
    return eAB_success;
}
```

b-kurczynski:
Incorporated the following changes:

1. Moved calls to `S3SyncSampleVolumeAndPan()` and `S3SyncSampleRate()` to follow call to `AudioBackend_PlaySample()`. This is required because the later one initializes the object on which `S3Sync...()` functions operate.
2. Removed a call to `ma_volume_db_to_linear()` as it expects input volume to be in exponential scale where `volume_db` is in linear scale. The call has been replaced with explicit transformation using reverse formula of the one found in `S3SyncSampleVolumeAndPan()` function.

Fixes #399

### DIFF
--- a/src/S3/s3sound.c
+++ b/src/S3/s3sound.c
@@ -350,9 +350,7 @@ int S3SyncSampleVolumeAndPan(tS3_channel* chan) {
     float pan_ratio; // [esp+38h] [ebp-8h]
     float total_vol; // [esp+3Ch] [ebp-4h]
 
-    int volume_db;
     int pan;
-    float linear_volume;
 
     if (chan->type != eS3_ST_sample) {
         return 1;
@@ -362,12 +360,8 @@ int S3SyncSampleVolumeAndPan(tS3_channel* chan) {
         total_vol = 1.0f;
     }
     if (chan->descriptor && chan->descriptor->type == chan->type) {
-        volume_db = 510.0f / total_vol * -5.0f - 350.0f;
-        if (volume_db >= 0) {
-            volume_db = 0;
-        }
 
-        if (AudioBackend_SetVolume(chan->type_struct_sample, volume_db) == eAB_success && chan->spatial_sound) {
+        if (AudioBackend_SetVolume(chan->type_struct_sample, total_vol) == eAB_success && chan->spatial_sound) {
 
             if (chan->left_volume != 0 && chan->right_volume > chan->left_volume) {
                 pan_ratio = chan->right_volume / (float)chan->left_volume;
--- a/src/harness/audio/miniaudio.c
+++ b/src/harness/audio/miniaudio.c
@@ -27,6 +27,9 @@ static int kMem_S3_DOS_SOS_channel = 234;
 typedef struct tMiniaudio_sample {
     ma_audio_buffer_ref buffer_ref;
     ma_sound sound;
+    int init_volume;
+    int init_pan;
+    int init_new_rate;
     int initialized;
 } tMiniaudio_sample;
 
@@ -158,6 +161,12 @@ tAudioBackend_error_code AudioBackend_PlaySample(void* type_struct_sample, int c
     }
     miniaudio->initialized = 1;
 
+    if (miniaudio->init_volume > 0) {
+        AudioBackend_SetVolume(type_struct_sample, miniaudio->init_volume);
+        AudioBackend_SetPan(type_struct_sample, miniaudio->init_pan);
+        AudioBackend_SetFrequency(type_struct_sample, rate, miniaudio->init_new_rate);
+    }
+
     ma_sound_set_looping(&miniaudio->sound, loop);
     ma_sound_start(&miniaudio->sound);
     return eAB_success;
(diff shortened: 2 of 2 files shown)


src/S3/s3sound.c
```diff
@@ -286,9 +286,6 @@ int S3PlaySample(tS3_channel* chan) {
         return 0;
     }
 
-    S3SyncSampleVolumeAndPan(chan);
```
jeff-1amstudios, line 289: These calls cannot be moved, as this is reverse engineered code and represents how the original code was structured. 

Only code inside `src/harness` can generally be changed.

In this case, take a look at `tMiniaudio_sample` struct. Perhaps we could add two new fields in that struct for volume and pan, and inside `AudioBackend_PlaySample` after calling `ma_sound_init_from_data_source`, we set the volume and panning to the values stored in the  `tMiniaudio_sample` struct.

src/harness/audio/miniaudio.c
```diff
@@ -182,8 +182,12 @@ tAudioBackend_error_code AudioBackend_SetVolume(void* type_struct_sample, int vo
     miniaudio = (tMiniaudio_sample*)type_struct_sample;
     assert(miniaudio != NULL);
 
-    // convert from directsound -10000 - 0 decibel volume scale
-    linear_volume = ma_volume_db_to_linear(volume_db / 100.0f);
+    // convert from Carmageddon "decibel" scale to linear <0.0f..1.0f>
+    linear_volume = (volume_db + 350.0f) / -5.0f;
```
jeff-1amstudios, line 186: this is one of these tricky parts where we are currently correctly emulating the behavior of the win95 directsound implementation. 

However, the win95 code was ported from the dos version with some unintended changes, and we are ok with changing the s3 code to more closely match the DOS version instead of the win95 version.

For example, https://github.com/dethrace-labs/dethrace/blob/main/src/S3/s3sound.c#L365-L368 does not appear in the DOS version at all. If it helps to remove it and just pass the raw `total_vol` to `AudioBackend_SetVolume`, that would be ok too, if it makes the resulting audio match up better with the dos behavior

b-kurczynski: What about https://github.com/dethrace-labs/dethrace/blob/main/src/S3/s3sound.c#L361-L363?

Does it exist in DOS version? It looks like avoidance of "divide by zero" in [L365-L368](https://github.com/dethrace-labs/dethrace/blob/main/src/S3/s3sound.c#L365-L368) and might not be needed too.

Nonetheless, if L365-L368 is removed, `total_vol` is in range <0..510> and would have to be scaled down to <0.0f..1.0f> either in `miniaudio.c` or `s3sound.c`. I guess a preference is to do it `miniaudio.c`, right?

jeff-1amstudios: Yes, in miniaudio.c

src/harness/audio/miniaudio.c
```diff
@@ -175,15 +184,19 @@ int AudioBackend_SoundIsPlaying(void* type_struct_sample) {
...
     assert(miniaudio != NULL);
 
-    // convert from directsound -10000 - 0 decibel volume scale
-    linear_volume = ma_volume_db_to_linear(volume_db / 100.0f);
+    if (!miniaudio->initialized) {
+        miniaudio->init_volume = volume;
+        return eAB_success;
+    }
+
+    linear_volume = volume / 510.0f;
```
jeff-1amstudios, line 199: @b-kurczynski with these changes, does the spatial audio sound to you like the dos original game?

dethrace-labs (approved):

PR #406, opened 2024-08-07
Ensure initial volume for CD music is properly set.

b-kurczynski:
First intention was to do a fix in a similar way like it has been fixed for effects volume. However, the "set-volume" request does not reach `S3SetCDAVolume()` function where caching of requested volume level could take place because it's stopped at `if (c->active)` [>>here<<](https://github.com/dethrace-labs/dethrace/blob/main/src/S3/s3.c#L1242).

As the fix enforced a change in original code, I decided to put it in a place where it's easier to understand. Other option was to do a fix in mentioned spot of S3.c file together with an implementation of deferred volume set in `S3SetCDAVolume()`, but that would change two files compared to one in proposed solution.

### DIFF
--- a/src/DETHRACE/common/sound.c
+++ b/src/DETHRACE/common/sound.c
@@ -699,6 +699,11 @@ int DRS3StartCDA(tS3_sound_id pCDA_id) {
                     }
                     gLast_tune = pCDA_id;
                     gCDA_is_playing = DRS3StartSoundNoPiping(gMusic_outlet, pCDA_id);
+#if defined(DETHRACE_FIX_BUGS)
+                    // Not a bug, but this is the less abusing place
+                    // where the initial music volume might be set with success.
+                    DRS3SetOutletVolume(gMusic_outlet, 42 * gProgram_state.music_volume);                    
+#endif
                     gCDA_tag = gCDA_is_playing;
                     if (!gCDA_is_playing) {
                         gCD_is_disabled = 1;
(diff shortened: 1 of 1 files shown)


dethrace-labs:
When you say "initial volume" do you mean the volume as set when the game starts and loads the music volume from config file, or when changing the music volume from the sound options screen?

b-kurczynski:
I meant the game can't set a proper volume to the music as it stored in the config.
Let say you set the Music Volume to 0 in the Settings menu, quit the game, and then start again. Music will be at full volume.

b-kurczynski:
Hi @dethrace-labs , this PR hangs for some time already. Should I do something more or there are objections to let it go to main?

Serasul1984:
its crazy to me that this over 10 years old bug, you find in many carmageddon games that also happen during map loading can be fixed now.
This gives me hope that other very annoying bugs from the original game can be fixed too, like "sound of car engines gets silent when to many things explode or crash at the same time"

dethrace-labs (approved):

src/DETHRACE/common/sound.c
```diff
@@ -699,6 +699,11 @@ int DRS3StartCDA(tS3_sound_id pCDA_id) {
                     }
                     gLast_tune = pCDA_id;
                     gCDA_is_playing = DRS3StartSoundNoPiping(gMusic_outlet, pCDA_id);
+#if defined(DETHRACE_FIX_BUGS)
+                    // Not a bug, but this is the less abusing place
```
dethrace-labs, line 703:
```suggestion
                    // Initial CD music volume was not set correctly
```

dethrace-labs:
Thanks for the ping! LGTM, made a suggestion to more clearly document the original behavior bug

PR #408, opened 2024-08-11
Nintendo 3DS port

MrHuu:
This is a very basic initial port to the 3DS platform.

- BRender software rendering
- no miniaudio implementation
- runs relatively slow on New3DS

It builds using the 3DS.cmake toolchainfile provided by devkitpro:
cmake .. -DCMAKE_TOOLCHAIN_FILE=/opt/devkitpro/cmake/3DS.cmake


tpimh:
Great work! I believe PSP port would take a bit more effort, but it's definitely possible.

Rinnegatamante:
@MrHuu You don't have Issues enabled on your fork so leaving it here: https://github.com/Rinnegatamante/dethrace/blob/main/src/harness/audio/miniaudio.c

Here I've implemented both a native SceAudio and an SDL2 audio backend. The SDL2 one should be usable for your 3DS port.

MrHuu:
Thank you very much @Rinnegatamante ! I was able to create a quick build using your SDL2 implementation and it's working quite nice. There is some stuttering during the intro videos, but in the menu's and in-game it seems to work perfectly. I'll update the 3DS port to reflect these changes.

Nice work on your Vita port, i very much appreciate your work towards both SDL2 and Vita implementations. Thanks again for pointing this out!

MrHuu:
I ended up in recreating a native ndsp audio backend, based on the SceAudio implementation. Increased the sample amount to fix the stuttering during the videos. And while at it, also re-implemented some of the other fixes from the Vita port.

Overall i was able to gain a few fps, and the working audio and music really makes the difference.

Thank you very much @Rinnegatamante !

Rinnegatamante:
You're welcome. The thread stack increase by the way, don't know if it's necessary. I just keep it for good norm on my ports toa void any silly edge case stack overflow.

MrHuu:
Yeah, i havn't checked some of the memory allocations properly yet. It probably has to be finetuned a bit, as i already noticed the audio from the smacker videos consume a lot of heap. They currently run out of memory about half way causing the audio to stop during playback. Reducing the stacksize and libctru allocated linear heap seems to fix it.

PR #414, opened 2024-09-25
Polish localization support

BSzili:
This PR adds support for the Polish localized version of Carmageddon released in 2002 as a covermount disc for the Click! magazine. I added a wrapper to isalnum to deal with extended ASCII characters that can cause problems for the parser.

### DIFF
--- a/src/DETHRACE/common/utility.c
+++ b/src/DETHRACE/common/utility.c
@@ -8,6 +8,7 @@
 #include "globvrpb.h"
 #include "graphics.h"
 #include "harness/config.h"
+#include "harness/hooks.h"
 #include "harness/trace.h"
 #include "input.h"
 #include "loading.h"
@@ -292,7 +293,7 @@ char* GetALineWithNoPossibleService(FILE* pF, unsigned char* pS) {
         if (ch != -1) {
             ungetc(ch, pF);
         }
-    } while (!isalnum(s[0])
+    } while (!Harness_Hook_isalnum(s[0])
         && s[0] != '-'
         && s[0] != '.'
         && s[0] != '!'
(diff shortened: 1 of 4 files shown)


dethrace-labs (approved):

dethrace-labs:
thanks!

MaxGracz:
I'm still having trouble with the Polish fonts. I tried changing English to Polish in path.txt, but it didn't help.

path
fonts

BSzili:
Not sure if direct linking is allowed, but it's on archive.org under the name "carmageddon-pl". If you search for carmageddon and set the language to Polish results only you'll also find it. This is the only version I tested.

madebr:
> I have iso of orginal click version. I can upload it for You somewhere

That won't be necessary anymore. BSzill pointed me in the right direction.
Screenshot for what's expected to be displayed (for both the game head-up, and the message shown when enabling/disabling mirrors) would be useful.
Also, can you decode the files using the `IWANTTOFIDDLE` cheat, and share a readable `DATA/TEXT.TXT` here?

(All using original executables that show correct headups)

MaxGracz:
> If the original exe also shows wrong glyphs in hires mode, it means this game data bug always existed. You'll have to replace `DATA/64X48X8/FONTS/NEWHITE.TXT` and/or `DATA/64X48X8/FONTS/NEWHITE.PIX` with fixed version. Fixing these glyphs in dethrace for the Polish language specifically is out-of-scope for dethrace.

So it's not for me. I don't know much about it...

> extended ASCII characters

Can you show me exactly what tables these are? Because I read that the ASCII tables do not contain Polish characters. 
My version alo works on old deathrace exe, so this support becomes unnecessary....

PR #427, opened 2024-12-12
Fixed random rotation in Wrecks Gallery on first click

b-kurczynski:
When player wants to look closer for a certain car in post race gallery (Wrecks Gallery) its model rotates by random angle when LMB is clicked. This change fixes the issue which is certainly a typo done during reverse engineering of code.


dethrace-labs (approved):
Thanks!

PR #428, opened 2024-12-16
Fix APC/BIGAPC wheels when player wasted equivalent cop car in previous race

b-kurczynski:
This is a fix to a bug which existed in original game. If player wasted a cop car driving a cop car then in next race player's APC wheels do not rotate nor turn in any direction. The same applies in case Suppressor is wasted by Suppressor.

### DIFF
--- a/src/DETHRACE/common/structur.c
+++ b/src/DETHRACE/common/structur.c
@@ -480,6 +480,10 @@ void DoGame(void) {
     int second_select_race;
     int first_summary_done;
     int i;
+#if defined(DETHRACE_FIX_BUGS)
+    // Player's APC/BIGAPC wheels got stuck if the equivalent cop car was wasted in previous race
+    int power_up_levels[3];
+#endif
     LOG_TRACE("()");
 
     gAbandon_game = 0;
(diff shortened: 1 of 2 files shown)


src/DETHRACE/common/structur.c
```diff
@@ -523,6 +528,28 @@ void DoGame(void) {
                     SwapNetCarsLoad();
                 }
             } else {
+#if defined(DETHRACE_FIX_BUGS)
+                // Player's APC/BIGAPC wheels got stuck if the equivalent cop car was wasted in previous race
+                if(strcmp(gProgram_state.car_name, "APC.TXT") == 0 ||
+                   strcmp(gProgram_state.car_name, "BIGAPC.TXT") == 0) {
+                    AboutToLoadFirstCar();
```
dethrace-labs, line 535: I don't think this is required, or the calls to switch resolution, or SetCarStorageTexturingLevel.

could you try without these?

b-kurczynski: That was my first approach but it didn't look good
image

The HUD is completely misaligned and cockpit view is in lowres even when hires is enabled. After that I decided to repeat complete "science & magic" that happens on regular car change and that started to work.

PR #431, opened 2025-01-05
ci: use SDL2 from head

madebr:
SDL2 from head includes a patch that does not fail CMake configuration when libX11 can be found, but no xext.h header.

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -48,7 +48,7 @@ jobs:
           cmake-arguments: ${{ matrix.platform.cmake-args }}
           cmake-generator: Ninja
           cmake-toolchain-file: ${{ matrix.platform.cmake-toolchain-file }}
-          version: 2-latest
+          version: 2-head
           add-to-environment: true
 
       - name: 'Prepare sources for release'


madebr:
All 2.x releases from the SDL2 are ABI compatible, so breakage is no issue.

b-kurczynski (approved):

Squall-Leonhart:
> All 2.x releases from the SDL2 are ABI compatible, so breakage is no concern.

breakage is always a concern, 

bind to a release tag, and not get surprised by accidental regressions you can't track for.

madebr:
> > All 2.x releases from the SDL2 are ABI compatible, so breakage is no concern.
> 
> breakage is always a concern,
> 
> bind to a release tag, and not get surprised by accidental regressions you can't track for.

True, but I trust the SDL devs on this one. ;)

b-kurczynski:
So what exactly happen? If I see correctly the runner switched from Ubuntu 22.x to 24.x, right? If there is a risk of regression due to usage of SDL2's bleeding-edge, and the issue is that 24.x is missing certain header files (missing packages?) compared to 22.x, then maybe it's better to revert back to use latest SDL2 and just install the missing package to 24.x like I did attempt in my PR?

Squall-Leonhart:
> True, but I trust the SDL devs on this one. ;)

you should probably look at the SDL 2.30.6 regressions then.

madebr:
> > True, but I trust the SDL devs on this one. ;)
> 
> you should probably look at the SDL 2.30.6 regressions then.

What regressions? Can you link a SDL GitHub issue?

For dethrace on Linux ci, since we don't ship a SDL2 elf library in the artifacts, we only care about the API/ABI interface, which is frozen.

madebr:
> 24.x does not contain libxss1 thus no libxext6 thus no `Xext.h` header files? There is even a suggestion to install missing package here: https://github.com/dethrace-labs/dethrace/actions/runs/12359607398/job/34492714851?pr=428#step:6:421

For Linux ci purposes, we don't really need x11 or wayland support. The only thing we care about on Linux is running the ci headless tests, and the dethrace executable correctly linked against a valid SDL2 library.
If we really  want wayland/x11 support, we'd need [a lot of extra development packages](https://github.com/libsdl-org/SDL/blob/main/docs/README-linux.md#build-dependencies), of which libxext-dev is only one.
The `libsdl-org/setup-sdl` has a [`install-linux-dependencies` option](https://github.com/libsdl-org/setup-sdl/blob/0b4058be1a69a43f9875e9a17f41e636495f5852/action.yml#L40-L41) which should automatically install all dependencies.

b-kurczynski:
> > 24.x does not contain libxss1 thus no libxext6 thus no `Xext.h` header files? There is even a suggestion to install missing package here: https://github.com/dethrace-labs/dethrace/actions/runs/12359607398/job/34492714851?pr=428#step:6:421
> 
> For Linux ci purposes, we don't really need x11 or wayland support. The only thing we care about on Linux is running the ci headless tests, and the dethrace executable correctly linked against a valid SDL2 library. If we really want wayland/x11 support, we'd need [a lot of extra development packages](https://github.com/libsdl-org/SDL/blob/main/docs/README-linux.md#build-dependencies), of which libxext-dev is only one. The `libsdl-org/setup-sdl` has a [`install-linux-dependencies` option](https://github.com/libsdl-org/setup-sdl/blob/0b4058be1a69a43f9875e9a17f41e636495f5852/action.yml#L40-L41) which should automatically install all dependencies.

Correct. However, please notice that until the runner was on 22.x the x11-ext packages where there and everything worked ok with no side effects. What is proposed now is: 1) use SDL2's head instead of official version; 2) SDL2 head IMO contains questionable fix to make Dethrace CI working - it alters CMake checks behavior and might create regression for the others and no longer provides valuable error/warning message.

b-kurczynski:
> > 1. use SDL2's head instead of official version;
> 
> Correct. That is the C in CI.
> 
> > 2. SDL2 head IMO contains questionable fix to make Dethrace CI working - it alters CMake checks behavior and might create regression for the others and no longer provides valuable error/warning message.
> 
> Right. SDL3 fails when there is no X11 or no Wayland support (there is a CMake option to allow this). Perhaps SDL2 should do the same.

The issue seems to be solely on Dethrace ground. CI/CD environment has changed because Ubuntu 24.x offers different set of pre-installed packages than 22.x. To me it means that the fix should be applied in Dethrace repo only. What is proposed is [>>this<<](https://github.com/libsdl-org/SDL/commit/16ad3e5af6ac12af6d6e06c49d78f4d6ec081519) change which might spawn a wave of new issues to the others - I would call it potential domino effect. I do understand that Dethrace does not use X11, but that's the prerequisite for SDL2.

It's all I can tell to convince you to change the proposed solution and more importantly to revert the change pushed to SDL2 repo. It's up to you which way you pick.

dethrace-labs:
I added https://github.com/dethrace-labs/dethrace/pull/432 as the smallest change required to fix the github build

PR #442, opened 2025-03-27
SDL3 migration

Link4Electronics:
- The cursor stopped working, but the game works fine.

### DIFF
--- a/cmake/FindSDL.cmake
+++ b/cmake/FindSDL.cmake
@@ -1,0 +1,242 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
+# file Copyright.txt or https://cmake.org/licensing for details.
+
+#[=======================================================================[.rst:
+FindSDL
+-------
+
+Locate the SDL library
+
+
+Imported Targets
+^^^^^^^^^^^^^^^^
+
+.. versionadded:: 3.19
+
+This module defines the following :prop_tgt:`IMPORTED` target:
+
+``SDL::SDL``
+  The SDL library, if found
+
+Result variables
+^^^^^^^^^^^^^^^^
+
+This module will set the following variables in your project:
+
+``SDL_INCLUDE_DIRS``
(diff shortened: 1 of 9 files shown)


madebr:
About SDL3, I think we want to merge Jeff's 3dfx pr first.
Also, I think it is worthwile to investigate ways to share SDL code between SDL2/3.
Copy/pasting SDL2/3 code leads to code rot.
This would also be useful for supporting SDL1, which runs on old platforms (DOS, Windows 9x).

cmake/FindSDL.cmake
```diff
@@ -0,0 +1,242 @@
+# Distributed under the OSI-approved BSD 3-Clause License.  See accompanying
```
madebr, line 1: This file can be removed. It is part of CMake.

src/S3/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ target_include_directories(s3
         include
 )
 
-target_link_libraries(s3 PRIVATE brender SDL2::SDL2 harness compile_with_werror)
+target_link_libraries(s3 PRIVATE brender SDL3::SDL3 harness compile_with_werror)
```
madebr, line 8: We need to better separate s3 and other helper libraries from SDL2.

We still want to support SDL2.
Not all platforms support SDL3 currently, or will be ported to SDL3.
e.g. UWP support ended in upstream SDL3, which is useful for Xbox Series homebrew. There's also an unofficial Switch port that does not yet have a SDL3 port. (Though, if you have an official Nintendo Switch SDK, and are under an NDA, you can get an "official" SDL3 port)

Link4Electronics:
> About SDL3, I think we want to merge Jeff's 3dfx pr first.

I was following https://wiki.libsdl.org/SDL3/README/migration not necessarily copying and pasting, but better wait then, I'll try to collaborate on the project somehow

PR #444, opened 2025-04-01
Support multiple platform drivers

madebr:
This makes it possible to build one dethrace executable that supports multiple platforms.
(Platform in this context means a gui platform: glfw, sdl2, win32, macos, ...)

Based on this, support for SDL1 (sdl 1.2 actually :nerd_face: ) is added.
Because symbols of sdl1 and sdl2 clash (they share identical names), these are loaded dynamically.

Running cmake with default argument does not change the current behavior: sdl2 is still the one and only supported platform, and is linked against.

It is only when you configure with `-DDETHRACE_PLATFORM_SDL1=ON` and/or `-DDETHRACE_PLATFORM_SDL_DYNAMIC=ON` that you will get SDL1 support and/or dynamically loading of SDL.

This pr depends on https://github.com/dethrace-labs/BRender-v1.3.2/pull/14

`dethrace` built with this pr and https://github.com/dethrace-labs/BRender-v1.3.2/pull/14 will result in an executable that only depends on libc, even though it supports SDL1 and SDL2.
```
$ ldd dethrace
	linux-vdso.so.1 (0x00007f3028506000)
	libm.so.6 => /usr/lib64/libm.so.6 (0x00007f302841f000)
	libc.so.6 => /usr/lib64/libc.so.6 (0x00007f302823d000)
	/lib64/ld-linux-x86-64.so.2 (0x00007f3028508000)
```


src/harness/platforms/sdl12.c
```diff
@@ -0,0 +1,236 @@
...
+    return screen;
+}
+
+static int set_window_pos(void* hWnd, int x, int y, int nWidth, int nHeight) {
+    // SDL_SetWindowPosition(hWnd, x, y);
+    if (nWidth == 320 && nHeight == 200) {
+        nWidth = 640;
+        nHeight = 400;
+    }
+    SDL_UpdateRect(hWnd, x, y, nWidth, nHeight);
```
arczi84: This is causing garbage on screen after exit on AmigaOS. Is it needed?

madebr: Perhaps the real issue we're nowhere calling `SDL_DestrowWindow` (or `SDL_FreeSurface` in SDL1)?

I created this using sdl12-compat on Linux. I'm going to try "real" sdl1 now.

madebr: I think this issue got introduced by ca1ef76e7f2beef062394b2b16c2d5149c668c5f:

`PDShutdownSystem` of  `pc-win95/win95sys.c` more complete, but it got disabled by this commit:
https://github.com/dethrace-labs/dethrace/blob/2e31fc6956a5a10598d6afe023af8bbedc3278ba/src/DETHRACE/pc-win95/win95sys.c#L1

madebr: I think this is a regression introduced by the 3dfx patch, causing `DestroyWindow` not being called.
I've created a ticket at https://github.com/dethrace-labs/dethrace/issues/446

madebr: Since this is not a regression, I'd like to address #446 in a later pr.
I've got something local that addresses #446, but I'd like this one merged first.

madebr:
The last commits add ci support for SDL1.

(While reviewing, don't forgot that this pr depends on https://github.com/dethrace-labs/BRender-v1.3.2/pull/14)

madebr:
Just a reminder to say this pr is ready for final review: https://github.com/dethrace-labs/BRender-v1.3.2/pull/14 got merged in the BRender repo

dethrace-labs (approved):

PR #454, opened 2025-05-16
Add SDL3 platform driver

madebr:


dethrace-labs (approved):

.github/workflows/workflow.yaml
```diff
@@ -51,7 +51,8 @@ jobs:
           cmake-generator: Ninja
           cmake-toolchain-file: ${{ matrix.platform.cmake-toolchain-file }}
           discriminator: ${{ matrix.platform.arch }}
-          version: 2-latest
+          version: 3-latest
+          version-sdl2-compat: 2-head
           version-sdl12-compat: 1-head
```
madebr, line 56: This builds SDL3, sdl2-compat and sdl12-compat.

SDL1 is EOL since 2012, SDL2 is in maintenance mode.
So it makes sense to change the order of platforms to SDL3 first in the near future, long before a next release.

madebr:
Jeff, can you re-review?

PR #460, opened 2025-06-17
Annotate all globals and functions with their address in carm95 executable

dethrace-labs:
- Prepare for some matching-assembly antics :)

### DIFF
--- a/src/DETHRACE/common/brucetrk.c
+++ b/src/DETHRACE/common/brucetrk.c
@@ -16,6 +16,7 @@
 
 // GLOBAL: CARM95 0x53e488
 br_actor* gMr_blendy;
+
 // GLOBAL: CARM95 0x550244
 int gDefault_blend_pc;
 
--- a/src/DETHRACE/common/brucetrk.c
+++ b/src/DETHRACE/common/brucetrk.c
@@ -14,10 +14,13 @@
 #include <stdlib.h>
 #include <string.h>
 
+// GLOBAL: CARM95 0x53e488
 br_actor* gMr_blendy;
+// GLOBAL: CARM95 0x550244
 int gDefault_blend_pc;
 
 // IDA: void __usercall AllocateActorMatrix(tTrack_spec *pTrack_spec@<EAX>, br_actor ****pDst@<EDX>)
+// FUNCTION: CARM95 0x4a8a26
 void AllocateActorMatrix(tTrack_spec* pTrack_spec, br_actor**** pDst) {
     tU16 z;
     LOG_TRACE("(%p, %p)", pTrack_spec, pDst);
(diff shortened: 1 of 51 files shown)


madebr:
Perhaps also commit the `reccmp-project.yml` file so we know we're sure we're comparing the same file?
(don't forget to update `.gitignore` for `reccmp-user.yml` and `reccmp-build.yml`)

src/DETHRACE/common/brucetrk.c
```diff
@@ -14,10 +14,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+// GLOBAL: CARM95 0x53e488
 br_actor* gMr_blendy;
+
+// GLOBAL: CARM95 0x550244
 int gDefault_blend_pc;
 
 // IDA: void __usercall AllocateActorMatrix(tTrack_spec *pTrack_spec@<EAX>, br_actor ****pDst@<EDX>)
```
madebr, line 23: Are the IDA annotations still useful at this point?

dethrace-labs: in some cases they have been - any time I touch the DOS executables, with IDA at least its super helpful to be able to copy-paste the function definition. We do have them in the `dethrace/codegen-output` repo though, so maybe they aren't so required here any more

dethrace-labs: And yes, on the reccmp files :+1 

This was the required first step, next step is to get it all compiling with MSVC 4.2 (up till now I was just testing on some individual functions pasted into a MSVC 4.2 project).

Once thats done, will hook up reccmp as a github action. Then the actual hard work starts :)

madebr: Since you're currently deep into ida to increase the accuracy and the annotations are useful for you, let's leave them in then :)

madebr: > And yes, on the reccmp files :+1

Since we will (probably) not try to match BRender, you probably might want to add an extra header with addresses of these functions, and also crt, [like this](https://github.com/isledecomp/isle/blob/master/LEGO1/library_msvc.h).

src/DETHRACE/common/brucetrk.c
```diff
@@ -14,10 +14,14 @@
 #include <stdlib.h>
 #include <string.h>
 
+// GLOBAL: CARM95 0x53e488
```
madebr: Looks good. My only nitpick is I would have formatted the addresses as `0x%08x` instead of `0x%x`

src/DETHRACE/common/car.c
```diff
@@ -4657,6 +4857,7 @@ void ViewNetPlayer(void) {
 }
 
 // IDA: void __cdecl ViewOpponent()
+// FUNCTION: CARM95 0x00487248
 void ViewOpponent(void) {
     static int n;
```
madebr, line 4876: These local global values also need annotations.

PR #465, opened 2025-07-28
Support for compiling with MSVC 4.20 for binary matching

dethrace-labs:


madebr (approved):
This is missing ci, but is looking good to me.

Issue #475, opened 2025-09-17
segfault on Fedora 43 / gcc 15.2.1


MastaG:
I've compiled dethrace per instructions and I've copied my gamefiles `~/.local/share/dethrace`:
```
ls ~/.local/share/dethrace/
3DFX_DOS.BDD  CARMAG.EXE    CARMID03.HMI  DETHRSRC.SYM  HMIDET.386   MELODIC.BNK  SETUP.INI     SOFTH.BDD   SOFTIZBF.BDD  SOFTT.BDD   VOODO2C.exe
3DFX_FAQ.TXT  CARMAGV.EXE   CONNECT.BAT   DOS4GW.EXE    HMIDRV.386   MUSIC        SMACKW32.DLL  SOFTHZ.BDD  SOFTIZF.BDD   SOFTTZ.BDD
ABUSE.TXT     CARMAV.EXE    CONNSOFT.BAT  DOSINST.BD    HMIMDRV.386  S3.INI       SNDSETUP.EXE  SOFTI.BDD   SOFTIZFF.BDD  SPLAT.DAT
CARM3DFX.BAT  CARM_FAQ.HTM  CREDITS.TXT   DRUM.BNK      INT          SCREAM2.WAV  SOFTB.BDD     SOFTIF.BDD  SOFTPRMF.BDD  SPLAT.GOG
CARMA.EXE     CARM_FAQ.TXT  DATA          glide2x.ovl   MCGA.BDD     SETUP.EXE    SOFTD.BDD     SOFTIZ.BDD  SOFTRNDF.BDD  VESA.BDD
```

However when I try and run the game it seems to crash instantly.
The backtrace doesn't seem to tell me much:

```
(gdb) thread apply all bt full

Thread 3 (Thread 0x7fffe37e26c0 (LWP 39219) "dethrace"):
#0  0x00007ffff7cd2982 in __syscall_cancel_arch () from /lib64/libc.so.6
No symbol table info available.
#1  0x00007ffff7cc6c3c in __internal_syscall_cancel () from /lib64/libc.so.6
No symbol table info available.
#2  0x00007ffff7cc72ac in __futex_abstimed_wait_common () from /lib64/libc.so.6
No symbol table info available.
#3  0x00007ffff7cc997e in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libc.so.6
No symbol table info available.
#4  0x00007ffff7411bcc in ?? () from /lib64/libEGL_nvidia.so.0
[...]

PR #476, opened 2025-09-17
Avoid fclose(NULL) and make [FI]ReturnBetween return in-range numbers

madebr:
This fixes 2 issues:
- avoid calling `fclose` with a NULL file pointer
- fix `[FI]RandomBetween` returning out-of-range numbers

Fixes https://github.com/dethrace-labs/dethrace/issues/475


dethrace-labs (approved):

dethrace-labs:
thanks!

PR #479, opened 2025-09-23
loadsave.c - cap number of vehicles saved

yukichigai:
Add a limit of 48 vehicles to save stolen status, as writing more than this will corrupt the save file and cause errors when it is loaded.


src/DETHRACE/common/loadsave.c
```diff
@@ -638,7 +638,7 @@ void MakeSavedGame(tSave_game** pSave_record) {
     for (i = 0; i < gNumber_of_races; i++) {
         (*pSave_record)->race_info[i].been_there_done_that = gRace_list[i].been_there_done_that;
     }
-    for (i = 0; i < gNumber_of_racers; i++) {
+    for (i = 0; i < gNumber_of_racers && i < 48; i++) { // Tracking more than 48 vehicles corrupts save files. This only matters in modded games that add more stealable vehicles
```
madebr: Fixes relative to OG should be guarded with `#ifdef DETHRACE_FIX_BUGS`.
```suggestion
#ifdef DETHRACE_FIX_BUGS
// Tracking more than 48 vehicles corrupts save files.
// This only matters in modded games that add more stealable vehicles
#define VALID_SAVE_GAME_INDEX(SAVE_GAME, INDEX) ((INDEX) < gNumber_of_racers && (INDEX) < COUNT_OF((SAVE_GAME)->opponent_info))
#else
#define VALID_SAVE_GAME_INDEX(SAVE_GAME, INDEX) ((INDEX) < gNumber_of_racers)
#endif
    for (i = 0; VALID_SAVE_GAME_INDEX(*pSave_record, i); i++) {
#undef VALID_SAVE_GAME_INDEX
```

yukichigai: Tested that and it seems to work. A lot more elegant than my change.

madebr (approved):
To avoid incompatibility with dethrace and/or the original games, please modify `CalcLSChecksum` to make it incompatible (in your forks).
This can be done easily by changing the initial value of `checksum` to another non-zero value:
https://github.com/dethrace-labs/dethrace/blob/d5bdca5edaa30eac11b20918fe69274cf2508e60/src/DETHRACE/common/loadsave.c#L109

PR #484, opened 2025-09-25
Opponent matching

dethrace-labs:


src/DETHRACE/CMakeLists.txt
```diff
@@ -61,6 +61,8 @@ endif()
 
 target_sources(dethrace_obj PRIVATE
     common/skidmark.c
+    common/opponent.c
```
madebr, line 64: Is there significance to the change in order?

src/DETHRACE/common/opponent.c
```diff
@@ -254,8 +254,6 @@ void ProcessCurrentObjective(tOpponent_spec* pOpponent_spec, tProcess_objective_
     case eOOT_wait_for_some_hapless_sod:
         ProcessWaitForSomeHaplessSod(pOpponent_spec, pCommand);
         break;
-    case eOOT_rematerialise:
-        break;
```
madebr, line 258: Can these be surrounded with `#ifdef DETHRACE_FIX_BUGS`?
My compiler emits a warning when a switch is missing a enum value.

src/DETHRACE/CMakeLists.txt
dethrace-labs: yes, this is starting to follow the ordering of compilation units in the original binary. The first 9 functions of `skidmark` are perfect matches, so are all aligned in the same memory locations as the original.

Unfortunately, as soon as one function is slightly larger or smaller all the others go out of alignment. Still, ideally we'd have the same ordering of functions and eventually same alignment

PR #486, opened 2025-09-25
Add Arch Linux AUR link to `README.md`

buzz:
Add a note in `README.md` pointing Arch Linux users to the AUR package: https://aur.archlinux.org/packages/dethrace


madebr:
My 2c is to not add this to the readme as this does not scale.
Instead, let's create a wiki page where users can advertise their ports/distributions etc.

PR #493, opened 2025-10-01
WIN32: replace LOG_DEBUG() with LOG_DEBUG2()

carlo-bramini:
Compiling dethrace rises this error when it is compiled with MinGW:

```
dethrace/src/harness/os/windows.c: In function 'OS_GetAdapterAddress':
dethrace/src/harness/os/windows.c:440:47: error: macro 'LOG_DEBUG' passed 2 arguments, but takes just 1
  440 |         LOG_DEBUG("name: %s", aa->FriendlyName); // Skip if name is provided and doesn't match FriendlyName
      |                                               ^
```

According to the content of `dethrace/src/harness/include/harness/trace.h`, I replaced with `LOG_DEBUG2()` macro and the problem has been solved.

### DIFF
--- a/src/harness/os/windows.c
+++ b/src/harness/os/windows.c
@@ -437,7 +437,7 @@ int OS_GetAdapterAddress(char* name, void* pSockaddr_in) {
     }
 
     for (IP_ADAPTER_ADDRESSES* aa = adapter_addrs; aa != NULL; aa = aa->Next) {
-        LOG_DEBUG("name: %s", aa->FriendlyName); // Skip if name is provided and doesn't match FriendlyName
+        LOG_DEBUG2("name: %s", aa->FriendlyName); // Skip if name is provided and doesn't match FriendlyName
         if (wcslen(wideName) > 0 && wcscmp(aa->FriendlyName, wideName) != 0)
             continue;
 


madebr (approved):
I can reproduce, and this fixes the error in `src/harness/os/windows.c`

carlo-bramini:
> I also see a fatal error in `src/DETHRACE/pc-all/allnet.c`, which probably needs casting.

I noticed it too, but perhaps there is a better solution for this than doing type casting.
POSIX standard declares `setsockopt` and `recvfrom` as:
```
int setsockopt(int socket, int level, int option_name, const void *option_value, socklen_t option_len);
ssize_t recvfrom(int socket, void *restrict buffer, size_t length, int flags, struct sockaddr *restrict address, socklen_t *restrict address_len);
```
MSDN declares them as:
```
int setsockopt(SOCKET s, int level, int optname, const char *optval, int optlen);
int recvfrom(SOCKET s, char *buf, int len, int flags, sockaddr *from, int *fromlen);
```
Actually, we can see these main differences:

1. option is a `const void *` into POSIX while MSDN declares it as `const char *`.
2. the length is a `socklen_t` into POSIX, while MSDN declares it as `int`.

In my opinion, there is a quick way for fixing the first item.
All we need to do is to add these lines to CMakeLists.txt:
```
if (MINGW)
    target_compile_options(dethrace PRIVATE -Wno-error=incompatible-pointer-types)
endif()
```
it will print a warning but it won't rise an error anymore.

About the second item, in my opinion you just need to declare `sa_len` as `socklen_t` and it is solved too.
MSDN declares the functions with an `int` parameter but winsock provides also `socklen_t` which is a `typedef` to `int` and so it's compatible without warnings.

PR #494, opened 2025-10-01
Fix demo compatibility

madebr:
Regression introduced in 81116169cfefa549a70a57f3c7e616c57b697973

### DIFF
--- a/src/DETHRACE/common/utility.c
+++ b/src/DETHRACE/common/utility.c
@@ -98,6 +98,14 @@ void EncodeLine(char* pS) {
     FILE* test;
     unsigned char c;
 
+#ifdef DETHRACE_FIX_BUGS
+    // Demo has its own decryption key + behavior
+    if (harness_game_info.mode == eGame_carmageddon_demo) {
+        EncodeLine_DEMO(pS);
+        return;
+    }
+#endif
+
     len = strlen(pS);
     key = (char*)gLong_key;
     if (gEncryption_method == 0) {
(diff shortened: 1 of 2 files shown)


dethrace-labs:
Ooops, thanks for this, my bad. I'll have to be more careful in the future. 

Confirmed this does fix loading demo assets, but it does not fix the (existing broken) `I AM FIDDLING` text file decryption that was reported in Discord. 
That uses `EncodeLine2` which needs to be pulled from the demo disassembly (`EncodeLine` was different in the demo and the retail code).

Since they are two different problems, happy to

dethrace-labs:
> Since we never added extra functionality, I'm inclined to ignore those cheats for the demo.
Normally, I'd agree. I guess in this case though, since some people will be playing with the demo content without access to the retail content, it would be great for IWANTTOFIDDLE to allow them to decrypt the files so they can play around with them - add new cars etc. The only other option is for someone to dig up an ancient copy of `CV.EXE` which can decrypt them

madebr:
Alternatively, we add demo-support to `[decode_datatxt.py`](https://github.com/dethrace-labs/dethrace/blob/main/tools/decode_datatxt.py).

dethrace-labs (approved):

dethrace-labs:
> Alternatively, we add demo-support to [decode_datatxt.py](https://github.com/dethrace-labs/dethrace/blob/main/tools/decode_datatxt.py).

And also the spidering into subfolders and skipping of txt files that shouldn't be decrypted (eg `KEYMAP`s etc)

madebr:
I think `IWANTTOFIDDLE` now fully works with the demo data files.
I tested as following:

This script pulls a file 3 times through the decoder, and verified whether the diff is zero:
```sh
#!/bin/sh

tools/decode_datatxt.py --method demo $1 >pass1.txt
tools/decode_datatxt.py --method demo pass1.txt > pass2.txt
tools/decode_datatxt.py --method demo pass2.txt > pass3.txt

diff -u pass1.txt pass3.txt
```

PR #500, opened 2025-10-05
SDL3: hide cursor

BeWorld2018:
Simple error


madebr (approved):
Thanks for finding and fixing this bug!

PR #514, opened 2025-10-23
TestBigEndian got deprecated, so better endian check

Link4Electronics:
with this and the patch for BRender-v1.3.2 repo, tested on big and little-endian machines, this matter should be done now.

### DIFF
--- a/src/DETHRACE/common/racestrt.c
+++ b/src/DETHRACE/common/racestrt.c
@@ -1560,11 +1560,19 @@ void SelectRaceDraw(int pCurrent_choice, int pCurrent_mode) {
     if (*test) {
         test2 = *test;
     }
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -85,7 +85,6 @@ function(add_compile_flags_if_supported TARGET)
     endforeach()
 endfunction()
 
-#test_big_endian(IS_BIGENDIAN)
```
madebr: We need to bump the minimum CMake version to 3.20 for this to be available by spec.

madebr: (which is fine by me)

Link4Electronics: Done.

dethrace-labs (approved):

Issue #417, opened 2024-10-15
Game crashes after winning race when you try to press P to bring back the 


fluddsskark:
Steps to recreate:

- Make sure that you had already disabled the facecam using P during the race.
- Press P immidiately after completing the race, and then it crashes the application.

dethrace-labs:
Thanks for the report

PR #532, opened 2025-12-23
Fix pratcam null pointer dereference

madebr:
Fixes https://github.com/dethrace-labs/dethrace/issues/417

### DIFF
--- a/src/DETHRACE/common/pratcam.c
+++ b/src/DETHRACE/common/pratcam.c
@@ -13,6 +13,7 @@
 #include "loading.h"
 #include "pd/sys.h"
 #include "sound.h"
+#include "structur.h"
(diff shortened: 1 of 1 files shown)


src/DETHRACE/common/pratcam.c
```diff
@@ -13,6 +13,9 @@
 #include "loading.h"
 #include "pd/sys.h"
 #include "sound.h"
+#ifdef DETHRACE_FIX_BUGS
```
dethrace-labs: we don't know what the header file inclusions looked like, so i don't think we need to wrap an ifdef here

madebr: Indeed. I got rid of it.

PR #537, opened 2025-12-24
Fix some float precision diffs

disinvite:
Non-exhaustive search for diffs where the only fix required is to flip the precision used in the float constant or expression.


src/DETHRACE/common/crush.c
```diff
@@ -245,7 +245,7 @@ void CrushModelPoint(tCar_spec* pCar, int pModel_index, br_model* pModel, int pC
             old_vector = *target_point;
             for (bend_axis = 0; bend_axis < 3; bend_axis++) {
                 target_point->v[bend_axis] += (1.0f - the_neighbour->factor / 256.0f) * movement.v[bend_axis];
-                
+
```
disinvite, line 248: clang-format fix

madebr (approved):

dethrace-labs (approved):

PR #538, opened 2025-12-25
Do not timeout main menu when cutscenes are disabled

madebr:
When starting the game with `-nocutscenes`, the main menu is fading out and re-appearing every 30s.
This pr disables the timeout when the game is started with `-nocutscenes`.

### DIFF
--- a/src/DETHRACE/common/structur.c
+++ b/src/DETHRACE/common/structur.c
@@ -62,6 +62,13 @@ tU32 gLast_checkpoint_time;
 // GLOBAL: CARM95 0x00551dbc
 tRace_over_reason gRace_over_reason;
 
+#ifdef DETHRACE_FIX_BUGS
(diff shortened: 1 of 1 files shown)


src/DETHRACE/common/structur.c
```diff
@@ -706,7 +706,15 @@ void DoProgram(void) {
             if (gGame_to_load >= 0) {
                 DoLoadGame();
             } else {
-                DoMainMenuScreen(30000u, 0, 0);
+#ifdef DETHRACE_FIX_BUGS
```
dethrace-labs: what about pulling this define up to the top of the file, or maybe in `const.h` where we've put some others? Not sure which one is "better"

madebr: I pulled `MAIN_MENU_TIMEOUT` to the top of `structur.c`

PR #540, opened 2025-12-29
Improve `DoWheelDamage`

disinvite:
Should be 100%. The remaining diffs start here:

```diff
@@ 0x4bf98f @@
mov dword ptr [ebp - 0x1c], 0        (crush.c:1302)
jmp 0x3
inc dword ptr [ebp - 0x1c]
-mov eax, dword ptr [ebp - 0x1c]
-cmp dword ptr [gNum_active_cars (DATA)], eax
-jle 0x3c6
+mov eax, dword ptr [gNum_active_cars (DATA)]
+cmp dword ptr [ebp - 0x1c], eax
+jge 0x3c6
mov eax, dword ptr [ebp - 0x1c]      (crush.c:1303)
```

because the `cmp` instruction is 1 byte smaller in recomp.

I took out this bit to improve accuracy:
```cpp
default:
    TELL_ME_IF_WE_PASS_THIS_WAY();
    break;
```
If those should stay in, let me know.

### DIFF
--- a/src/DETHRACE/common/crush.c
+++ b/src/DETHRACE/common/crush.c
@@ -1298,70 +1298,48 @@ void DoWheelDamage(tU32 pFrame_period) {
     br_vector3 wonky_vector;
     static int kev_index[4];
 
-    if (gAction_replay_mode && ReplayIsPaused()) {
-        return;
-    }
-
-    for (i = 0; i < gNum_active_cars; i++) {
-        car = gActive_car_list[i];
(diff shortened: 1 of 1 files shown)


madebr:
> I took out this bit to improve accuracy:
> 
> ```c++
> default:
>     TELL_ME_IF_WE_PASS_THIS_WAY();
>     break;
> ```
> 
> If those should stay in, let me know.

I think that's fine.
It depends on the size of the `tCar_spec::wheel_dam_offset` array, which should always be 4.

dethrace-labs (approved):

PR #554, opened 2026-01-25
World.c matching

dethrace-labs:
All functions in `world.c` either matching or pushed closer to matching

### DIFF
--- a/src/DETHRACE/common/world.c
+++ b/src/DETHRACE/common/world.c
@@ -23,6 +23,7 @@
 #include "pedestrn.h"
 #include "piping.h"
 #include "replay.h"
+#include "skidmark.h"
 #include "spark.h"
 #include "trig.h"
 #include "utility.h"
@@ -1828,6 +1829,12 @@ int FindSpecVolIndex(br_actor* pActor) {
     int i;
     tSpecial_volume* v;
 
+#ifdef DETHRACE_FIX_BUGS
+    if (pActor == NULL) {
+        return -1;
+    }
+#endif
+
     for (i = 0, v = gProgram_state.special_volumes; i < gProgram_state.special_volume_count; i++, v++) {
         if (gSpec_vol_actors[i] == pActor) {
             return i;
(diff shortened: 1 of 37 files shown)


src/DETHRACE/common/world.c
```diff
@@ -265,28 +269,28 @@ void DisposeStorageSpace(tBrender_storage* pStorage_space) {
 void ClearOutStorageSpace(tBrender_storage* pStorage_space) {
     int i;
 
-    for (i = 0; pStorage_space->pixelmaps_count > i; ++i) {
-        if (pStorage_space->pixelmaps[i] != NULL) {
+    for (i = 0; i < pStorage_space->pixelmaps_count; i++) {
+        if (pStorage_space->pixelmaps[i]) {
```
madebr: Does the removal of `!= NULL` have an effect on accuracy?

src/DETHRACE/common/world.c
```diff
@@ -301,81 +305,86 @@ void ClearOutStorageSpace(tBrender_storage* pStorage_space) {
...
-    for (i = 0; i < pStorage_space->pixelmaps_count; i++) {
-        if (pStorage_space->pixelmaps[i]->identifier
-            && ((br_pixelmap*)pThe_pm)->identifier
-            && strcmp(pStorage_space->pixelmaps[i]->identifier, ((br_pixelmap*)pThe_pm)->identifier) == 0) {
-            return eStorage_duplicate;
+    if (pStorage_space->pixelmaps_count < pStorage_space->max_pixelmaps) {
+        for (i = 0; i < pStorage_space->pixelmaps_count; i++) {
+            if (pStorage_space->pixelmaps[i]->identifier
+                && ((br_pixelmap*)pThe_pm)->identifier
+                && strcmp(pStorage_space->pixelmaps[i]->identifier, ((br_pixelmap*)pThe_pm)->identifier) == 0) {
```
madebr: Shouldn't we rather change the prototype of `AddPixelmapToStorage` to accept a `br_pixelmap* pThe_pm` instead?

src/DETHRACE/common/world.c
dethrace-labs: thanks, didn't mean to remove the explict NULL check

src/DETHRACE/common/world.c
dethrace-labs: yes, good point. I took a look at the codegen, it did generate the `br_pixelmap**` correctly based on the debug symbols, but likely it was an oversight in the original code and should have been `br_pixelmap*` the whole time. Or there is some bug in codegen, who knows. But I've changed it

PR #570, opened 2026-02-16
ci: retry cpack a few times

madebr:
DMG creation on macOS occasionally fails, so retry a few times

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -118,7 +118,6 @@ jobs:
            echo "Package creation failed after $max_tries attempts."
            exit 1
           fi
-          cpack .
       - name: Upload Artifact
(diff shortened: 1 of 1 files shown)


.github/workflows/workflow.yaml
```diff
@@ -101,6 +101,23 @@ jobs:
...
+            break
+           fi
+           echo "Package creation failed. Sleep 1 second and try again."
+           sleep 1
+          done
+          if test $success = 0; then
+           echo "Package creation failed after $max_tries attempts."
+           exit 1
+          fi
           cpack .
```
dethrace-labs: we can remove the final `cpack` now, since we did it already on line 110?

madebr:
```suggestion
```

It was an extra one for good measure :)

PR #575, opened 2026-02-16
Improve `CreateShrapnelShower`

disinvite:
Matching except for swapped stack order of the variables `vel` and `tv`. Also the instructions for this line:
```cpp
ts2 = BrVector3Dot(pNormal, &vel);
```
You can reorder `vel` at the top of the function to get this to match.

### DIFF
--- a/src/DETHRACE/common/spark.c
+++ b/src/DETHRACE/common/spark.c
@@ -795,9 +795,9 @@ void CreateShrapnelShower(br_vector3* pos, br_vector3* v, br_vector3* pNormal, b
     }
     ts2 = pNormal->v[1] * ts;
 
-    tv.v[0] = v->v[0] - ts2 * pNormal->v[0];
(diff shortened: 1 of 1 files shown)


madebr (approved):

dethrace-labs:
> You can reorder vel at the top of the function to get this to match.

the ordering of stack variables should be (almost always) correct - the ordering comes from the symbol dump. The symbol dump comes from a pre-release 3dfx splatpack DOS build. The only times the stack hasn't matched that I've seen are when a variable was added as part of 3dfx support - for example https://github.com/dethrace-labs/dethrace/blob/c08202bb0de85c58a1d913d25c041936f4680415/src/DETHRACE/common/loading.c#L618-L622
In that example `result` was not in the carm95 build, so I ifdef'd it out. Those are very minor exceptions though.

When `vel` and `tv` are swapped, I don't change the stack, I change the usage so `vel = BrVector3Dot(pNormal, &ts2);`

PR #579, opened 2026-02-22
Bump BRender and pipe BRender errors to log

dethrace-labs:
- Adds missing call to `PDInstallErrorHandlers` from the win95 code
- Routes the failure function to `LOG_PANIC` which writes the message to `dethrace.log` before exiting

This should fix the issue we have on Windows where the default BRender `null` dialog handler is used, which swallows errors and allows execution to continue, leading to the infamous "white screen in 3dfx" mode behavior :)


dethrace-labs:
> all I get is the Windows abort handler (=dialog with abort/retry/ignore).

And you will also see the error in `dethrace.log` file regardless of the diagnostics or not.

PR #586, opened 2026-02-27
Add Windows-only cli arg to select gpu accelerator

madebr:
The default is to select the discrete GPU over the CPU integrated in the CPU.

Running the game with `--gpu-accelerator` or `--no-gpu-accelerator` allows me to select NVIDIA accelerator or the Intel CPU integrated graphics.

### DIFF
--- a/src/harness/harness.c
+++ b/src/harness/harness.c
@@ -24,6 +24,11 @@ extern const tPlatform_bootstrap SDL1_bootstrap;
 extern const tPlatform_bootstrap SDL2_bootstrap;
 extern const tPlatform_bootstrap SDL3_bootstrap;
 
+#ifdef _WIN32
+__declspec(dllexport) unsigned long NvOptimusEnablement                  = 1;
+__declspec(dllexport) unsigned long AmdPowerXpressRequestHighPerformance = 1;
+#endif
+
 static const tPlatform_bootstrap* platform_bootstraps[] = {
 #if defined(DETHRACE_PLATFORM_SDL3)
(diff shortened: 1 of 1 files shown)


madebr:
This only works for x86-64. A 32-bit dethrace always selects NVIDIA on my system.

src/harness/harness.c
```diff
@@ -444,6 +449,16 @@ int Harness_ProcessCommandLine(int* argc, char* argv[]) {
                 safe_strcpy(harness_game_config.platform_name, argv[i + 1]);
                 consumed = 2;
             }
+#ifdef _WIN32
+        } else if (strcasecmp(argv[i], "--gpu-accelerator") == 0) {
```
dethrace-labs, line 453: I wonder if the terminology will be confusingly similar to the opengl use hardware acceleration option.

How about something like `--force-discrete-gpu` or `--force-high-performance-gpu`?

dethrace-labs: Also, whats the use case for someone to set these flags?

madebr: > How about something like `--force-discrete-gpu` or `--force-high-performance-gpu`?

Since this does not work on 32-bit x86 for me, I think naming it a hint would be better. e.g. `--hint-performance-gpu` and `--hint-no-performance-gpu`

madebr: > Also, whats the use case for someone to set these flags?

The use case is to work around the wrong colors for 64-bit x86_64 Windows

PR #592, opened 2026-03-03
network.c matching

jeff-1amstudios:
```
Total effective accuracy 92.42% across 2376 functions (92.19% actual accuracy)
```

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -3,9 +3,9 @@
 Dethrace looks for a `dethrace.ini` file in the [system preferences directory](https://wiki.libsdl.org/SDL2/SDL_GetPrefPath):
 | Platform | Example Path                                                                 |
 |----------|----------------------------------------------------------------------|
-| Windows  | `C:\Users\bob\AppData\Roaming\dethrace\dethrace.ini`                |
-| macOS    | `/Users/bob/Library/Application Support/dethrace/dethrace.ini`      |
-| Linux    | `/home/bob/.local/share/dethrace/dethrace.ini`                      |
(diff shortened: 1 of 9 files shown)


src/DETHRACE/pc-none/nonet.c
```diff
@@ -68,6 +68,7 @@ int BroadcastMessage(void) {
 }
 
 // IDA: int __cdecl PDNetInitialise()
+// FUNCTION: CARM95 0x004543F8
```
madebr: Don't these annotations belong in `src/DETHRACE/pc-win95/win95net.c` instead?
(Perhaps annotate them as stubs?)

dethrace-labs: thanks, I was being lazy since `win95net` wasn't being used to compile for msvc 4.2.

I fixed it today so that `nonet.c` is removed, and `win95net.c` is now used correctly for matching, which means we can put the annotations there.

src/DETHRACE/common/network.c
```diff
@@ -870,96 +869,137 @@ int NetSendMessageToAllPlayers(tNet_game_details* pDetails, tNet_message* pMessa
...
-        return 4;
+        the_size = 4;
+        break;
+    }
+    return the_size;
+
+    if (the_size > 255) {
+        char s[256];
+        sprintf(s, " type %d", pType);
+        FatalError(kFatalError_NetContentsTooBig_S, s);
```
madebr, line 1003: Is this really needed for matching?

dethrace-labs: yep! I guess it was an original bug that moved the return statement above the check. Or maybe they moved it deliberately late in development when they were satisfied the check was no longer needed 🤷

PR #736, opened 2026-03-27
DRS3OutletSoundsPlaying 100% match

jeff-1amstudios:
## Match result

```
0x46493a: DRS3OutletSoundsPlaying 100% match.

✨ OK! ✨
```

#### Original match

```
---
+++
@@ -0x46493a,6 +0x4aa996,11 @@
0x46493a : push ebp 	(sound.c:373)
0x46493b : mov ebp, esp
0x46493d : push ebx
0x46493e : push esi
0x46493f : push edi
0x464940 : -cmp dword ptr [gSound_enabled (DATA)], 0
         : +call abort (FUNCTION) 	(sound.c:374)
         : +pop edi 	(sound.c:375)
         : +pop esi
         : +pop ebx
         : +leave 
         : +ret 


DRS3OutletSoundsPlaying is only 58.82% similar to the original, diff above
```

*AI generated. Time taken: 387s, tokens: 235,018*

### DIFF
--- a/src/S3/s3.c
+++ b/src/S3/s3.c
@@ -1224,12 +1224,10 @@ int S3OutletSoundsPlaying(tS3_outlet* pOutlet) {
     if (!gS3_enabled) {
         return 0;
     }
-    c = pOutlet->channel_list;
-    while (c) {
+    for (c = pOutlet->channel_list; c != NULL; c = c->next) {
         if (c->active) {
             sound_count++;
         }
-        c = c->next;
     }
     return sound_count;
 }
--- a/src/S3/s3.c
+++ b/src/S3/s3.c
@@ -1215,6 +1215,25 @@ int S3StopSound(tS3_sound_tag pTag) {
     return 0;
 }
 
+// FUNCTION: CARM95 0x0049C6E2
+int S3OutletSoundsPlaying(tS3_outlet* pOutlet) {
+    int sound_count;
+    tS3_channel* c;
+
+    sound_count = 0;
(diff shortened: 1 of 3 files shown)


src/S3/s3.c
```diff
@@ -1215,6 +1215,25 @@ int S3StopSound(tS3_sound_tag pTag) {
...
+    if (!gS3_enabled) {
+        return 0;
+    }
+    c = pOutlet->channel_list;
+    while (c) {
+        if (c->active) {
+            sound_count++;
+        }
+        c = c->next;
+    }
```
madebr: If accuracy is identical, a for loop would make it fit better with the current code style.
```suggestion
    for (c = pOutlet->channel_list; c != NULL; c = c->next) {
        if (c->active) {
            sound_count++;
        }
    }
```

src/S3/s3.c
```diff
@@ -1215,6 +1215,25 @@ int S3StopSound(tS3_sound_tag pTag) {
     return 0;
 }
 
+// FUNCTION: CARM95 0x0049C6E2
+int S3OutletSoundsPlaying(tS3_outlet* pOutlet) {
```
jeff-1amstudios, line 1219: wow it created this function and decompiled it by itself

madebr: How come ci is failing? Perhaps your scripts are not pushing all changes?

Serasul1984:
I hope someone can fix this

dethrace-labs (approved):

PR #765, opened 2026-03-28
Draw smoke of opponent correctly when in cockpit mode

### LINKED ISSUE #523: Damaged car smoke fume offsets are anywhere near them but not on the bonnet
Image
Image

madebr:
Fixes #523

### DIFF
--- a/src/DETHRACE/common/spark.c
+++ b/src/DETHRACE/common/spark.c
@@ -1415,6 +1415,12 @@ void RenderSmoke(br_pixelmap* pRender_screen, br_pixelmap* pDepth_buffer, br_act
     tU32 seed;
     tU32 not_lonely;
 
+#ifdef DETHRACE_FIX_BUGS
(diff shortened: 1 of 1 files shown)


src/DETHRACE/common/spark.c
```diff
@@ -1415,6 +1415,12 @@ void RenderSmoke(br_pixelmap* pRender_screen, br_pixelmap* pDepth_buffer, br_act
     tU32 seed;
     tU32 not_lonely;
 
+#ifdef DETHRACE_FIX_BUGS
+    // SetWorldToScreen sets gCameraToScreen matrix, which is used by SmokeCircle3D
```
dethrace-labs, line 1419:
```suggestion
    // Fix opponent smoke rendering in cockpit view with rearview mirror enabled
    // SetWorldToScreen sets gCameraToScreen matrix, which is used by SmokeCircle3D
```

madebr: You'll have to sneak this change in with another pr.

PR #901, opened 2026-04-10
Remove unit tests and Unity test framework

jeff-1amstudios:
## Summary
- remove the entire `test/` directory
- remove Unity test framework sources and test harness files
- remove root CMake test wiring (`BUILD_TESTS` option and `add_subdirectory(test)`)

## Notes
- this PR fully removes unit test infrastructure from the repository


madebr:
Do we still need the null platform then?

madebr:
@jeff-1amstudios I fixed the fatal dethrace warnings. Can you have a another look at it?

PR #920, opened 2026-04-10
Disable werror in builds

dethrace-labs:
Before https://github.com/dethrace-labs/dethrace/pull/901, we weren't building properly with warnings-as-errors. Now we are, we have warnings for uninitialized things.

With the asm-match stuff going on, I'd rather keep this off to avoid lots of PRs failing, and do separate clean-up work to fix uninitialized vars etc

### DIFF
--- a/.github/workflows/workflow.yaml
+++ b/.github/workflows/workflow.yaml
@@ -14,13 +14,13 @@ jobs:
     strategy:
       matrix:
         platform:
-          - { name: "Linux", arch: "x64", os: ubuntu-latest, werror: true }
-          - { name: "Linux", arch: "arm64", os: ubuntu-24.04-arm, werror: true }
+          - { name: "Linux", arch: "x64", os: ubuntu-latest, werror: false }
+          - { name: "Linux", arch: "arm64", os: ubuntu-24.04-arm, werror: false }
           - {
               name: "MacOS",
               arch: "arm64-x64",
               os: macos-latest,
-              werror: true,
+              werror: false,
               cmake-args: '-DCMAKE_OSX_ARCHITECTURES="x86_64;arm64"',
             }
           - {


madebr:
From [this log](https://github.com/dethrace-labs/dethrace/actions/runs/24262006281/job/70847882889?pr=901), I believe the `-Werror=maybe-uninitialized` are a legitimate bug, and deserve to be fixed.
The `-Wformat-overflow` warning can be disabled, which I would do by passing `-Wno-format-overflow` in our CMake.

But I'd keep the `-Werror`.

dethrace-labs:
totally agree, i'm just proposing that we fix them in a separate stage to the asm matching. What I don't really want is a bunch of PRs that fail to build on CI because of original initialized variables which slows us down some.

versus dedicated run with `-Werror` and fix all the errors PRs. Once we're done with matching, make Werror default on again

madebr:
I don't 100% agree it will slow us down by a lot, just a little bit.
The problem appears because of a project change (=removal of unit tests) and once fixed, will only show for a few functions (I think).

As long as an issue is created as a reminder, I don't care if it's disabled (temporarily). MSVC users building the game in debug mode, will potentially see fatal errors.

dethrace-labs:
> once fixed, will only show for a few functions (I think).

what I'm worried about is all the functions that we are editing to potentially add usage of new uninitialized vars

madebr:
> what I'm worried about is all the functions that we are editing to potentially add usage of new uninitialized vars

Your llm can build dethrace locally with `-DDETHRACE_WERROR=OFF`, so it won't be bothered.
Since it creates a pr, we can evaluate what's needed for each pr.

I don't want to lower the standards for an llm.
We need to uphold a minimum level of quality.

PR #961, opened 2026-04-12
Fix DoColour256 and a few other things

madebr:
- Fix regression introduced in d8ff2397ec1ec50a3976f2f730e4d8a83e2b5dae
- I think `memmove` is at `0x004eaa10`
- Other small changes

### DIFF
--- a/src/DETHRACE/common/finteray.c
+++ b/src/DETHRACE/common/finteray.c
@@ -372,7 +372,7 @@ int DRModelPick2D__finteray(br_model* model, br_material* material, br_vector3* 
 // FUNCTION: CARM95 0x004abe8d
 int FindHighestPolyCallBack__finteray(br_model* pModel, br_material* pMaterial, br_vector3* pRay_pos, br_vector3* pRay_dir, br_scalar pT, int pF, int pE, int pV, br_vector3* pPoint, br_vector2* pMap, void* pArg) {
 
-    if ((pT) < gNearest_T) {
+    if (pT < gNearest_T) {
         gNearest_T = pT;
         gNearest_model = pModel;
         gNearest_face = pF;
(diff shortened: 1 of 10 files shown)


src/DETHRACE/common/mainloop.c
```diff
@@ -6,6 +6,7 @@
 #include "depth.h"
 #include "displays.h"
 #include "drmem.h"
+#include "errors.h"
```
madebr:
```suggestion
```

src/DETHRACE/common/finteray.c
```diff
@@ -372,7 +372,7 @@ int DRModelPick2D__finteray(br_model* model, br_material* material, br_vector3*
 // FUNCTION: CARM95 0x004abe8d
 int FindHighestPolyCallBack__finteray(br_model* pModel, br_material* pMaterial, br_vector3* pRay_pos, br_vector3* pRay_dir, br_scalar pT, int pF, int pE, int pV, br_vector3* pPoint, br_vector2* pMap, void* pArg) {
 
-    if ((pT) < gNearest_T) {
+    if (pT < gNearest_T) {
```
dethrace-labs: this is awkward but needed for matching by forcing the correct ordering. I'd rather (at this point) keep the higher match and allow for (just a little) awkwardness. Once we've matched as much as we can (98-99ish%, then we can decide, do we fork/branch from that spot, and from there, can be more open for quality of life changes, modernizing the code etc

madebr:
```suggestion
    if ((pT) < gNearest_T) {
```

madebr: I agree, as long as we keep the quality up.

dethrace-labs (approved):

PR #997, opened 2026-04-19
Fix buffer overflow in resolve_full_path

buzz:
The `resolve_full_path()` function in `src/harness/os/linux.c` and `src/harness/os/macos.c` uses a static buffer `_program_name[1024]`, but calls `getcwd(path, PATH_MAX)` where `PATH_MAX` is 4096 on Linux. This causes `getcwd()` to write up to 4096 bytes into a 1024-byte buffer, triggering a stack buffer overflow.

The crash is detected by glibc's `_FORTIFY_SOURCE=3` protection.

### DIFF
--- a/src/harness/os/linux.c
+++ b/src/harness/os/linux.c
@@ -194,21 +194,28 @@ static void signal_handler(int sig, siginfo_t* siginfo, void* context) {
     exit(1);
 }
 
-void resolve_full_path(char* path, const char* argv0) {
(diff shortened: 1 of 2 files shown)


dethrace-labs:
In addition to the change suggested here, should we also be defining _program_name with the right size macro to make sure we declare a buffer thats actually big enough?

buzz:
> In addition to the change suggested here, should we also be defining _program_name with the right size macro to make sure we declare a buffer thats actually big enough?

Good catch. Done in 267440c.

PR #1133, opened 2026-06-29
Fix build compilation on FreeBSD / BSD platforms

### LINKED ISSUE #1132: FreeBSD support
Hello all,

dethrace was ported to FreeBSD with very few changes bellow that upstream might consider to include.
Freebsd dethrace port is at: https://github.com/freebsd/freebsd-ports/tree/main/games/dethrace

Special note:

I did try sdl2 and sdl3 and found that sdl3 has more quality and graphics are more soft.
I'm using sdl3 by default and I'd like to know if upstream recommends using it at this moment, since sdl2 is cmake default.

Thanks for this great project!

```diff
--- src/DETHRACE/pc-all/allnet.c.orig	2026-06-28 21:41:50 UTC
+++ src/DETHRACE/pc-all/allnet.c
@@ -15,6 +15,7 @@
 #include <ctype.h>
 #include <stdlib.h>
 #include <string.h>
+#include <netinet/in.h>
 
 #ifdef _WIN32
 #define _WINSOCK_DEPRECATED_NO_WARNINGS
--- src/harness/CMakeLists.txt.orig	2026-06-28 21:33:35 UTC
+++ src/harness/CMakeLists.txt
@@ -72,7 +72,7 @@ elseif(APPLE)
     target_link_libraries(harness PRIVATE dbghelp ws2_32 iphlpapi)
 elseif(APPLE)
     target_sources(harness PRIVATE os/macos.c)
-elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux")
+elseif(CMAKE_SYSTEM_NAME STREQUAL "Linux" OR CMAKE_SYSTEM_NAME STREQUAL "FreeBSD")
     target_sources(harness PRIVATE os/linux.c)
 else()
     message(FATAL_ERROR "Unsupported or unknown platform: ${CMAKE_SYSTEM_NAME}")
--- src/harness/os/linux.c.orig	2026-06-28 21:37:35 UTC
+++ src/harness/os/linux.c
@@ -32,6 +32,7 @@
 #include <termios.h>
 #include <time.h>
 #include <unistd.h>
+#include <netinet/in.h>
 
 #if SUPPORT_TRACEBACK
 #include <execinfo.h>

```

nunotexbsd:
Regarding the addition of `<netinet/in.h>` in `allnet.c` and `os/linux.c`:

On POSIX-compliant systems (like FreeBSD and other BSD derivatives), types and constants related to internet addresses (such as `struct sockaddr_in` or macros like `IPPROTO_TCP` / `INADDR_ANY`) strictly require the inclusion of `<netinet/in.h>`. 

While on Linux these symbols are often implicitly pulled in through other network headers, on BSD systems they are not, which leads to compilation errors due to undefined types.

To ensure maximum portability across all UNIX systems without breaking Windows, we wrapped it under `#ifndef _WIN32` in `allnet.c` and placed it directly in `os/linux.c` (since that file is exclusively built for UNIX platforms now using the `UNIX` macro in CMake). This follows the POSIX standard and avoids cluttering the codebase with specific platform macros.

Closes #1132


src/DETHRACE/pc-all/allnet.c
madebr, line 32: Perhaps move the include here?

nunotexbsd: Done! I've moved the include inside the POSIX block and sorted it alphabetically as suggested. Thanks!

madebr (approved):

madebr:
Merged! Thanks for the issue and subsequent fix!

